Issue #16, opened 2014-12-03
Add a benchmark


mwestphal:
@christianbrinch @smaret 
Is there any way to test lime functionnality ?

I've been trying to compare results (image0.fits) but at each execution , the results are different.

is there a minimal example wich lime would resolve always the same way ?

smaret:
For reference, the original benchmark problems can be found [here](http://www.strw.leidenuniv.nl/astrochem/radtrans/).

PR #25, opened 2014-12-19
Python model files

smaret:
This branch implements Python model files. The model file, which defines the density, temperature, velocity and abundances, is read by LIME at run time. Other input parameters are read from a separate input file. This has several advantages over the previous implementation:

1. LIME does not need to be compiled for each model file
2. Model files are more easily written than in C
3. Other parameters are set from an input file rather than from a complicated C structure

In addition, the branch implements a CMake based build system. The code is built on a continuous built system (Travis CI) to make sure that the code compiles and runs fine.

This is major overhaul of LIME code. Before merging this PR into the dev branch, one needs to make sure that LIME still works as expected. For this we need to implement a test suite (see #16). The documentation also needs to be updated (see #26).


smaret:
Replaced by PR #33.

Issue #24, opened 2014-12-09
x[3] in photon.c:photon set but not used


mwestphal:
@smaret @christianbrinch 
in photon : https://github.com/smaret/lime/blob/a8711d11d88e5bbc9cc6a4ce44c41d8910c19165/src/photon.c#L188-191

x is set, but never used.

What is the purpose of it ? should it be removed ?

Issue #23, opened 2014-12-08
gsl accel alloc inccorect placement


mwestphal:
Provoked a memory leak.
fixed by : https://github.com/mwestphal/lime/commit/be1c6edd48cf210f0b6546e66e25406fbcb456d1

Issue #22, opened 2014-12-08
Memory bug in tcpsocket


mwestphal:
They were some memory error and leak in tcpsocket.c

fixed by : https://github.com/mwestphal/lime/commit/6fcda195410351d72bb89a30a0ef8dd2876094d6

Issue #20, opened 2014-12-04
rng not freed


mwestphal:
As in https://github.com/smaret/lime/issues/8, same in predefgrid

fixed by : https://github.com/mwestphal/lime/commit/f6b0fd7a47cf3797b30c9d6ccceb602714d9419f

PR #32, opened 2015-04-02
Fix memory leaks and overflows

smaret:
This PR fixes a number of memory leaks and buffer overflows in the code (#3, #4, #5, #7, #8, #9, #10, #11, #12, #13, #14, #15, #18, #20, #22, #23 and #24).


mwestphal:
that's an awesome push request ! :D

smaret:
@mwestphal: If there was a number-of-bug-fixed-per-PR contest at GitHub, you would win it easily.

PR #45, opened 2015-06-04
Bug fixes

allegroLeiden:
Here are a number of fixes to bugs which could affect the numerical output, but which are mostly of a minor character. The biggest change perhaps is that the order of channel maps in the output cube is now reversed.


smaret:
Should I close this or wait for definitive fix?

PR #50, opened 2015-07-07
Parallelized code

allegroLeiden:
There are a lot of commits here, but most of them are rearrangements to prepare the way for the final one, which allows the user to divide the bottleneck processing between several threads (via a new optional user-settable parameter 'nThreads').

The task now makes use of the OpenMP API provided by the gomp library to divide up between threads two sections of code: (i) the solution of the radiative transport equations; (ii) raytracing to make the final image.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,7 +42,7 @@ endif
 ##
 
 TARGET  = lime.x 
-CC		= gcc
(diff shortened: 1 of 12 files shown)


smaret:
Superseded by #56.

PR #53, opened 2015-07-15
No-curses option

smaret:
A rebased version of @imckstewart 's PR #51.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
 language: c
 before_install: sudo apt-get update
 install: sudo apt-get install libgsl0-dev libcfitsio3-dev libatlas-base-dev libncurses5-dev libqhull-dev libqhull5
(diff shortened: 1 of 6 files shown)


smaret:
@imckstewart : The completion % look incorrect. LIME says that the grid build is 99.9% completed when it starts, and then it takes a while before it is finished:

```
cd example
../lime -n model.c 
*** LIME, The versatile line modeling engine, Ver. 1.5
   Building grid: 99.90 percent done
```

Also the photon propagation completion cycles from 0% to 100% and back several times before completion.

smaret:
Note the `lime -n` to turn off ncurses output.

allegroLeiden:
I'll see if I can fix these things.

allegroLeiden:
Fixes in the latest nocurses upload.

PR #57, opened 2015-07-16
Do not rely on environment variables

smaret:
The lime script is modified to setup the environment variables needed to
compile and run the code (PATHTOLIME, WORKDIR and LD_LIBRARY_PATH)
internally. The only requirement is that the lime script is in the PATH.


smaret:
@imckstewart I'm thinking of including this in the upcomming 1.5 release. What do you think?

imckstewart:
I guess, if you want to. Some people might already have sourcing the sourceme in their .bashrc, but I guess even if they leave such a line in place, it will just append unnecessary extra copies to LD_LIBARY_PATH etc, which is pretty harmless.

Issue #58, opened 2015-07-16
Incorrect relation of dust to densities


allegroLeiden:
At (currently) line 69 of molinit.c the variable g[id].mol[s].knu[iline] is calculated. According to Michiel Hogerheijde, it is incorrect (or at least, leads potentially to an error in certain cirumstances) to derive this value from the density of the 0th collision partner as is currently done. However, it is not simply as simple as summing the densities of all coll partners, since some (I am told) play a greater role than others. Some thought (and discussion with the astronomers) needs to go into amending this. It is a bit late for the 1.5 release, but MH is happy to see it in a subsequent patch.

smaret:
I agree with Michiel. However, [as far as git remembers (Dec 2011)](https://github.com/lime-rt/lime/blob/52572a0254d1e917592b5cc25f0e912d8b47002b/src/molinit.c#L67), it has always been computed in this fashion in LIME. So this can probably wait for another release.

imckstewart:
Absolutely!

smaret:
Fixed with #128.

PR #59, opened 2015-07-16
Michiel fixes

allegroLeiden:
Michiel pointed out 2 problems, but only one can be quickly fixed. I raised the other as an issue (#58) for the future.


smaret:
The first commit of this PR (5f5ad51) has been already merged. Not sure why it is still showing up here. Also the commits 246847b and b1f643f looks identical. I'm cherry picking the first one as 187e2064d and closing this PR. Thanks for the fix!

imckstewart:
I made a spelling error in the first commit, did 'git commit --amend' to fix it, but that opened a can of worms it seems! I'm glad you found your way through the mess.

PR #62, opened 2015-08-15
Improved grid building and VTK output

attilajuhasz:
Dear Lime developers, 

My name is Attila Juhasz, currently I am a postdoc in the Institute of Astronomy in Cambridge. Previously I was a member of the dutch ARC node (Allegro) at the Leiden Observatory. While I was working for Allegro I have a made a few changes to Lime and I thought they might be useful for the community. I made a  few small changes to the VTK output (removing velocity normalisation and changing cell type), but the most significant change is in the grid building. I reduced the number of smoothing iterations in the Lloyd's algorithm from 20 to 5. This algorithm is regularising the grid which is not necessarily a good thing in all situations. Too much smoothing can seriously decrease the effect of weighted sampling in the grid point selection. I also moved the pointEvaluation() function from weights.c to the model.c as the weighting function might be edited by the users frequently. The final change to the grid building is that I introduced spatial regions. The point evaluation process during grid building runs for each spatial region separately. Each region (defined by 3D cartesian or spherical coordinate boundaries) has a certain number of grid points, a reference radius and a sampling type (i.e. logarithmic or linear). The weighting function can be also different for each region. Using such regions it is easy to add extra points to the grid to a very specific spatial location. This can be important if the molecular layer represents only a small fraction of the total model volume. The 'sampling' and 'pIntensity' in the parameter structure are omitted if regions are used. The sampling is now defined for each region separately, while the value of pIntensity is calculated as the total number of points in all regions together. 

Cheers
Attila

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -55,7 +55,7 @@ SRCS    = src/aux.c src/messages.c src/grid.c src/LTEsolution.c   \
 		  src/popsout.c src/predefgrid.c src/ratranInput.c      \
           src/raytrace.c src/smooth.c src/sourcefunc.c          \
 		  src/stateq.c src/statistics.c src/magfieldfit.c       \
-		  src/stokesangles.c src/writefits.c src/weights.c      \
+		  src/stokesangles.c src/writefits.c       \
 		  src/velospline.c src/getclosest.c  \
 		  src/tcpsocket.c src/defaults.c src/fastexp.c
 MODELS  = model.c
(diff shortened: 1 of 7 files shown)


smaret:
In the future, please submit a different PR for each new feature (e.g. one for the smoothing parameter, another one for the VTK format fix, etc.). We may not merge all the proposed changes, and it is harder to to if they all are in the same PR. Thanks!

attilajuhasz:
Here's the link to the VTK legacy format: http://www.vtk.org/wp-content/uploads/2015/04/file-formats.pdf
Sorry about the merged / rebased PR. I'll try to do separate ones in the future!

smaret:
Superseded by PR #70 and #71.

allegroLeiden:
We discussed separation of the grid density specification from the molecular density in the telecon in March 2016 but at that time I don't believe I had read this present thread (or had forgotten it). Anyway AFAIR in the telecon we agreed an action item on me to raise a PR with an extra user-supplied function to specify the grid point number density function, which should, to preserve backward compatibility, be optional and default to the 1.5 standard scheme. I agree that pointEvaluation() is not a particularly informative name - suggest something like gridDensity()? But anyway I guess this can be finessed/discussed after I make the PR.

PR #64, opened 2015-08-18
Save memory for sink points

ParfenovS:
Dear LIME team,

I've made relatively big number of changes in the LIME code version 1.4b. Most of these changes are for the better performance. I even made an Open MPI version of the code which works well and allows for heavy computations on HPC clusters. I'd like to contribute these changes (most of the them).

The commits which I'm sending now allow to safe memory. The code do not perform computations of level populations in the sink points. In the case of high number of molecular levels and grid points the allocated memory for the populations and collisional rates in the sink points is not used but can significantly consume RAM. I've made so that this unused memory is not allocated. According with this change I also changed the procedures of reading and writing of binary model files. At the moment changes in these two procedures are breaking because they do not allow to use binary files obtained with the current version of the LIME code. The compatibility can be saved by using dummy variables during reading/writing binary files but I've decided to not use such possibility because dummy variables will consume the space on a hard disk.

I'm really new with github and git. Please tell me if I made something wrong in the sense of technical things like commiting, pushing etc.

Regards

Sergey.

### DIFF
--- a/src/molinit.c
+++ b/src/molinit.c
@@ -256,7 +256,7 @@ molinit(molData *m, inputPars *par, struct grid *g,int i){
       }
     }
 
-    for(id=0;id<par->ncell;id++){
+    for(id=0;id<par->par->pIntensity;id++){
       g[id].mol[i].partner=malloc(sizeof(struct rates)*m[i].npart);
       for(ipart=0;ipart<m[i].npart;ipart++){
         g[id].mol[i].partner[ipart].up = malloc(sizeof(double)*m[i].ntrans[ipart]);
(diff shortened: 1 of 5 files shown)


smaret:
Thanks for the PR. How much RAM does it save for the example model?

ParfenovS:
Sorry for late answer.

 >How much RAM does it save for the example model?
It is about 20 MB or 30 %.

smaret:
I am not sure if 30% is sufficient to justify a change in the binary format. Are there specific cases for which you are limited by the memory footprint of LIME?

ParfenovS:
I used LIME for the computations with 20000 surface and 20000 grid 
points and with ~500 molecular energy levels. In such case the amount of 
memory that can be saved is measured in gigabytes. I understand that it 
is really specific case and it can be more important to save backward 
compatibility of the binary format.

01.10.2015 12:23, Sebastien Maret пишет:

smaret:
> I understand that it is really specific case and it can be more important to save backward compatibility of the binary format.

OK, I'm closing this PR then.

PR #65, opened 2015-08-20
Lapack/MKL support

ParfenovS:
These commits provide possibility to use LAPACK or Intel MKL linear equation solvers in the stateq procedure. Using of these solvers for the model provided in the example directory of the lime distribution do not lead to any significant changes in the results or time of the code execution (comparing with GSL solver). But if the file hco+@xpol.dat will be replaced on, for example, a-ch3oh.dat file from LAMDA database then there can be a problems with GSL solver. GSL solver can be very slow in that case or it can even fall with an error. LAPACK or MKL solvers allow to avoid such problems and allow to perform fast computations even when GLS fails (which often happens in the case of high number of molecular levels).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -71,7 +71,7 @@ MODELO 	= src/model.o
 
 #CCFLAGS = -O3 -falign-loops=16 -fno-strict-aliasing -DTEST -DFASTEXP -DNO_NCURSES
 CCFLAGS = -O3 -falign-loops=16 -fno-strict-aliasing
-LDFLAGS = -lgsl -lgslcblas -l${QHULL} -lcfitsio -lncurses -lm 
+LDFLAGS = -lgsl -lgslcblas -l${QHULL} -lcfitsio -lncurses -lm ${EXTRALDFLAGS}
 
 .SILENT:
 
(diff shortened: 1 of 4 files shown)


christianbrinch:
Hi Sergey
Thanks for your patches. Your contributions are much appreciated.
We are currently working on a way to evaluate and test submitted patches before we merge them into the main branch. Just so you know that we are not ignoring you. It might just take a little while before we merge your pull requests. 
Cheers,
Christian

ParfenovS:
Dear Christian,

Thank you for information!
I understand that checking and testing may take some time.

Regards

Sergey.

21.08.2015 12:05, Christian Brinch пишет:

smaret:
> But if the file hco+@xpol.dat will be replaced on, for example, a-ch3oh.dat file from LAMDA database then there can be a problems with GSL solver. GSL solver can be very slow in that case or it can even fall with an error.

Can you please provide an example `model.c` that demonstrate this problem?

ParfenovS:
Please find model.c in the attachment.

allegroLeiden:
I'm wondering if this issue (#85 etc) needs a bit further looking at. I am not convinced the best way forward is to throw a more highly-powered matrix solver at the problem. If the solver is having difficulties that would seem to me to be an indication of something non-optimal with the way the problem is being posed. I plan to look into this when time allows.

smaret:
> I am not convinced the best way forward is to throw a more highly-powered matrix solver at the problem.

I fully agree. In addition, the MKL library isn't free, and not all LIME users may it installed on their computer. So we can't really rely on it.

smaret:
PR #119 fixes #85 without the need for the MKL, so I'm closing this PR.

PR #66, opened 2015-08-22
Modification of predefgrid

ParfenovS:
The code (photon and traceray procedures) assumes that if the user supplies predefined grid then the velocity field is taken from this grid. The user may want to change the format of file with the predefined grid so that the velocity will not be taken from this file.

These commits allow the code to check whether the user supplies the velocity field with the predefined grid and to use appropriate functions in photon and traceray procedures. With these commits the code also will check whether other physical properties were taken from the input file with predefined grid. It may be useful if the user wants to change the format of the input file.

### DIFF
--- a/doc/usermanual.rst
+++ b/doc/usermanual.rst
@@ -409,6 +409,11 @@ If this option is used, LIME will not generate its own grid, but rather
 use the grid defined in this file. The file needs to contain all
 physical properties of the grid points, i.e., density, temperature,
 abundance, velocity etc. There is no default value.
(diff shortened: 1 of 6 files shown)


allegroLeiden:
I think this goes in the right direction, but I have felt all along that I/O for grid data must be made much more systematic, with a limited number of clearly defined possibilities for which grid attributes are expected to be supplied in a file submitted by the user. My PR #92 sets out such a scheme.

smaret:
> My PR #92 sets out such a scheme.

OK, I'm closing this one then

PR #70, opened 2015-09-03
VTK output fix

smaret:
This is PR contains @attilajuhasz's fixes to the VTK output in PR #62. The cell types codes in VTK format are documented [here](http://www.vtk.org/wp-content/uploads/2015/04/file-formats.pdf).


smaret:
@christianbrinch I need your inputs here. Do you remember by the velocity vector is normalized in the VTK output? Can we remove this normalization safely?

PR #72, opened 2015-09-09
Changed values of some constants defined in lime.h 

tlunttil:
The values (e.g. for speed of light) were slightly off from currently defined/experimentally determined SI values.

Added parentheses around preprocessor definitions to make sure that e.g. constant HCKB=100.*HPLANCK*CLIGHT/KBOLTZ is computed at compile time. This also makes the code slightly faster.

### DIFF
--- a/src/fastexp.c
+++ b/src/fastexp.c
@@ -12,6 +12,7 @@
 
 extern double EXP_TABLE_2D[128][10];
 extern double EXP_TABLE_3D[256][2][10];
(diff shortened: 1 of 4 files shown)


smaret:
Thanks for your contribution. Commits ee03e8a 718e220 and 6ffaf67 are not related to the constant definition though. Please open a new PR with the constant definition only. 

Also please remember to format your commits messages following our [contribution guidelines](CONTRIBUTING.md): the first line of 5dcf17a is too long and it is truncated in GitHub.

Issue #69, opened 2015-09-01
Raytracing progress bar appears to be broken


christianbrinch:
The progress bar used to track the ray-tracing progress. With version 1.5 it seems to stay at 0 until the ray-tracing is done (and then jump to 100%). 
I have only tried continuum ray-tracing so far, so I don't know if this bug also affects line ray-tracing too.

smaret:
I confirm that the bugs also affects the line ray-tracing.

@imckstewart: would you mind having a look at this?

allegroLeiden:
Fixed by 18bf031, branch message_fixes.

PR #75, opened 2015-09-17
Images in lines only, fix of bug#69

ParfenovS:
These commits fix the bug #69 and the bug in the writefits procedure.
Also these commits will allow the code to produce continuum-subtracted images.


smaret:
Bug #69 has already been fixed, so I can't merge this PR has it is. I'm closing it as invalid, but please feel free to open a new PR with your proposed changes to produce continuum-subtracted images.

PR #77, opened 2015-09-24
Inpars

allegroLeiden:
The intent of this is to separate the previous single inpars struct into two: one which contains ONLY those parameters the user can set, and the other for general program information (which may include some of the user pars). Besides using different variables to express different functionality, this makes it easier to change names for example without affecting the user interface. I had to make minor changes to code in a lot of places, but the result (i) compiles, (ii) passes valgrind tests, and (iii) produces an identical image to the previous.


allegroLeiden:
I guess you mean `ifndef`? Good idea though.

rschaaf-aifa:
O yes:

#ifndef LIME_H
#define LIME_H
.....
#endif /* LIME_H */

allegroLeiden:
I just force-pushed a commit which includes the macro test you proposed @rschaaf-aifa but undoes the move of the function prototypes to inpars.h. Hope I have not mangled the commit log too much.

allegroLeiden:
Serial conflict fixing seems to be unavoidable when the PRs have built up like this. I think we should try to be a bit more sequential with the next release.

PR #78, opened 2015-09-24
Message fixes

### LINKED ISSUE #69: Raytracing progress bar appears to be broken

allegroLeiden:
This adds a bit of missing functionality to the non-curses output, also fixes #69.


smaret:
Looks good, thanks!

PR #79, opened 2015-09-24
New raytrace

allegroLeiden:
This is a new optional algorithm which deals better with models which have areas where the Voronoi cells are much smaller than the image pixel size.


smaret:
Superseded by #145.

PR #80, opened 2015-09-24
Treegrid

allegroLeiden:
This is a new optional algorithm for generating grid points. The original algorithm is based on the rejection method, but this is slow when the (number) density function is highly peaked (i.e. when its maximum value is many times its average value). To compensate for this, the original algorithm flattens the user-input density function by raising it to a small positive power. An effect of this is that dense areas of the model are undersampled.

I ran this fairly extensive code addition/modification through valgrind to fix reported memory leaks etc. This addition seems to work ok in that it produces images broadly similar to the previous; it probably needs some more extensive testing though to investigate the effects of the new algorithm.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -49,24 +49,24 @@ endif
 ##
 
 TARGET  = lime.x 
-CC		= gcc -fopenmp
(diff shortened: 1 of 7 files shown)


smaret:
Superseded by #93.

PR #83, opened 2015-10-04
Small speed-ups and fixes to photon packet tracking

tlunttil:
Slightly faster FastExp Taylor approximation computation.

Fixes the step-length projection issue and the half-interval offset in segment calculation. Velocity projection turned out to be more difficult to fix, and it is not included in this pull request.


smaret:
Superseded by #112.

PR #76, opened 2015-09-24
Smallfixes

allegroLeiden:
Not much to say - spelling errors, order of lists, tweaks to constant values.

PR #84, opened 2015-10-13
Smallfixes

smaret:
This is a rebased version of Ian PR #76. All relevant commits should be listed here.


allegroLeiden:
@smaret can I push to ian-smallfixes?

src/lime.h
```diff
@@ -76,6 +76,7 @@
 #define N_RAN_PER_SEGMENT       3
 #define FAST_EXP_MAX_TAYLOR	3
 #define FAST_EXP_NUM_BITS	8
+#define N_SMOOTH_ITERS		20
```
smaret: @allegroLeiden Should we document this in the user manual? Or is it considered as an internal parameter?

allegroLeiden: My leaning would be 'internal', for now at least. With mods I plan for treegrid I am hoping to be able to dispense with the necessity to smooth altogether. With that in mind, I probably could have left the smooth iter setting alone - this change was just a software engineer's instinctive response when encountering a 'hard-wired' value in the body of the code. :-/

smaret: Agreed. I'm removing it from the release notes then.

Issue #86, opened 2015-11-13
Density calculation is incorrect when the first collision partner is not molecular hydrogen


kpolsen:
Hey there,

I'm trying to simulate [CII] line emission with LIME, and can't see a line profile!

So far, I only made the following edits to the example model, model.c, that comes with LIME:

hco+@xpol.dat -> CII.dat (see below)
img[0].unit			    = 0 -> 1
abundance[0] = 1.e-9 -> 1.4e-4 (in this case the CII/H2 ratio right? Set to be very high here, equal to the solar [C/H]/2 ratio)
abundance[0] = 50 (i.e. a constant Tk rather than one that varies with R)

For simplicity, I've only considered para-H2 molecules as collisional partners (even though [CII] can also be excited by collisions with electrons, HI and orto-H2). For that purpose, I reduced the LAMDA data file to: 
-------------------   CII.dat   ---------------------------
!MOLECULE
C+ (atomic ion)
!MOLECULAR WEIGHT
12.0
!NUMBER OF ENERGY LEVELS
2
!LEVEL + ENERGIES(cm^-1) + WEIGHT + J
    1     0.000000000   2.0     0.5  ! 2P
    2    63.395087      4.0     1.5  ! 2P
!NUMBER OF RADIATIVE TRANSITIONS
1
!TRANS + UP + LOW + EINSTEINA(s^-1) + FREQ(GHz) + E_u(K)
    1     2     1  2.300E-06        1900.5369        91.21
!NUMBER OF COLL PARTNERS
1
!COLLISIONS BETWEEN
2 C+ + pH2  ! thermal H2 J=0 and J=2, cf Wiesenfeld & Goldsmith 2014)
!NUMBER OF COLL TRANS
1
!NUMBER OF COLL TEMPS
7
!COLL TEMPS
   10.0   20.0   50.0  100.0  200.0  300.0  500.0 
!TRANS + UP + LOW + COLLRATES(cm^3 s^-1)
    1     2     1  4.36E-10 4.53E-10 4.63E-10 4.72E-10 5.13E-10 5.55E-10 6.01E-10
-------------------   CII.dat   ---------------------------

And the line profile that comes out is almost constant across all velocity bins!:

![cii1](https://cloud.githubusercontent.com/assets/13789517/11155755/bca071de-8a04-11e5-91b7-22a25de60dd0.png)

I must be doing something stupid. Hoping someone has time to help me out!

imckstewart:
[...]

PR #87, opened 2015-12-10
Fix for issue #86

ParfenovS:
The LIME code can not handle properly the situation when there is only one collision partner and it is not H2 (see issue #86 ). If the collison partner is para- or ortho-H2 the code will not compute molecular density. The pull request fixes this problem allowing for para- or ortho-H2 to be a single collison partner.


smaret:
Thanks for the fix, I'm merging it.

As noted in #91, we need to find a better way to provide the total H2 density at some point. Your comments on this issue are welcome!

smaret:
I've just noticed that your PR also contains two merge commits (#91715ca and #134748b) that shouldn't be there. Next time please remember to rebase your work onto the master branch to make sure that your PR only contains the relevant commits. For more information, see our guidelines.

Issue #88, opened 2016-02-02
Example model output


GaryFuller:
Hi, I have installed LIME and run the example which comes with the package which is an infalling envelope. The spectra in the output image have a redshifted peak which is stronger than the blueshifted peak. This is the opposite of what you expect from an infalling envelope. Any thoughts about what is happening?

imckstewart:
Hi Gary, this is a known issue offline, though nobody has yet raised it here. This looks a little bit like a simple sign error in the way lime processes velocities, but my own view is that there *was* a sign error in lime prior to the 1.5 release (see closed issue #38), which is now fixed, but which was masking and partially compensating a deeper problem which manifests itself as the reversed profile. I am trying to diagnose this when I get time to devote to it, but it will be a bit of a slow process I'm afraid. To assist in this I've written some code to allow full output of the grid point values at any stage in the code. I'll be integrating this with lime and making a pull request shortly. This will make it easier to see what the level populations look like along lines of sight which give this reversed profile. I can't rule out completely a simple sign error but the shifts in line centres if one uses a simple model with uniform v_z components all check out as expected, as does the hyperfine structure of lines which have this, so at this stage my own money is on some deeper error and not on something so simple as a reversed sign.

christianbrinch:
This issue stresses the importance of having a set of benchmark problems to test the code *all the time*.  My private copy which branched off around version 1.4 does not hav this issue.

GaryFuller:
Thanks. I'll be glad to hear when the problem is resolved.

imckstewart:
Christian have you tried adding, say, a constant 1e4 to your velocity z component and seeing which way the line shifts (for theta=phi=0). Benchmark problems and other test sets are great for when the code has moved out of the development (alpha) phase; I do not personally feel that lime has yet reached this status.

Issue #38, opened 2015-05-11
Possible velocity inversion


allegroLeiden:
Anika Schmiedeke (Uni Koeln) noted on about 20 Jan 2015:

I'm using LIME version 1.4beta (including the bugfix Christian implemented in 
the getjbar() function in the photon.c file in March 2014).

I initially only modeled SgrB2-M, so the peak of the lines were at 60 km/s. I 
then introduced SgrB2-N to the same model and expected the line peak towards 
SgrB2-N at ~64 km/s. But instead it was at ~56 km/s.

This confused me and I performed a lot of tests with the velocity. Since I'm 
using the predefgrid routine I simply set the gas velocities 

v_x = 0.0 km/s
v_y = 0.0 km/s
v_z = 50.0 km/s

everywhere. When using a source velocity of 60 km/s, the peak of the line was 
located at 70 km/s. When changing v_z to 70 km/s, the line peak was at 50 
km/s. This mirroring behaviour was consistent for all other velocities I 
tested. 

In the LIME paper, it is written, that LIME internally tracks photons by their 
frequency and not relative velocity (Section 2, at the end of page 2). Please 
correct me, if I'm wrong, but I figured, that the 
img[im].pixel[ipx].intense[ichan] array is ordered such that we have the 
frequency axis from left to right. 

So as a test, I've changed the output of the spectral axis to frequency:

	cdelt3  = img[im].velres * restfreq / CLIGHT;
	crpix3  = (double) img[im].nchan / 2. + 1;
	crval3  = restfreq * (1 - img[im].source_vel / CLIGHT);

	fits_write_key(fptr, TSTRING, "CTYPE3  ", &"FREQ    ",    "", &status);  
	fits_write_key(fptr, TDOUBLE, "CDELT3  ", &cdelt3,        "", &status);
	fits_write_key(fptr, TDOUBLE, "CRPIX3  ", &crpix3,        "", &status);
	fits_write_key(fptr, TDOUBLE, "CRVAL3  ", &crval3,        "", &status);
	fits_write_key(fptr, TSTRING, "CUNIT3  ", &"Hz      ",    "", &status);

Using Miriad to perform the switching to velocity (velsw task), everything was 
correct. The line peak for a gas velocity of 70 km/s was at 70 km/s etc. 

Instead of switching the output of the spectral axis to frequency, one could 
also just introduce a minus in the cdelt3 variable, which would invert the 
velocity axis and return the same result.

[...]

PR #90, opened 2016-02-16
Fix velocity inversion bug

allegroLeiden:
Fixes issue #88 in which the line shape of the default model seemed to be reversed (i.e. there was a red excess instead of the expected blue one). This problem had been masked before fixing issue #38.


smaret:
Thanks for the fix!

PR #92, opened 2016-02-19
Addition of facility to read and write the entire grid to FITS file.

allegroLeiden:
Much of the additional material in this PR is in the module grid2fits.c. In this module are comprehensive and tested routines to read and write the data in the list of grid structs from/to a FITS file. One problem with doing this is that the grid list contains different amounts of information at differing places in the code. It may not always be desired to wait the full processing time before outputting the grid - for example if the user just wants to look at the distribution of the points. For this reason the various scenarios have been divided into 4 stages in which grid contains differing amounts of information. The details of this are described in a long comment at the start of grid2fits.c.

For the time being I have included only some fairly rudimentary support for actually writing the grid (i.e., making use of the new functionality), and none at all for reading it. A more comprehensive service should probably wait on discussion about the future design of LIME's i/o.

17 June 2016: I reworked this PR substantially. I've now done the best job I can to (i) make the IO interface generic and not tied absolutely to FITS, and (ii) to implement fully read-and-write capability for the grid data. For the 2nd of these I have not yet been able to read the data at stage 4 (i.e. after all the level populations have values for them) and feed that either to further iterations of solver or direct to ray-tracing. Before that can be done, there needs to be a much more orderly succession of operations preparatory and during the level solving, which I felt was a bit beyond the scope of the present PR.


allegroLeiden:
As soon as this completes its Travis check it will be ready to merge. I'll do that tomorrow if nobody objects.

PR #93, opened 2016-02-26
Treegrid

smaret:
This is a rebased version of @imckstewart 's PR #80.


allegroLeiden:
I've merged the current master branch into this PR and, if there are no objections in the next day or so, I will merge it into master (i.e. accept the PR). A summary of the changes in this PR is as follows.

The PR addresses two separate challenges:
- A: choosing random points to follow some freely chooseable distribution function, and doing so efficiently (= quickly).
- B: making sure that no two points are arbitrarily close together.

A: The canonical algorithm for choosing randoms to follow some generic distribution is called the rejection method (RM). It is robust but can be very slow if the desired distribution function has large variations. LIME<=1.6 basically uses RM but applies a number of tweaks and biases designed to speed it up. These work well for the example model, but in fact only because they are *tuned* to suit the example model - for a different model they will be about as quick but will not necessarily produce anything close to an accurate sampling of the model.

The present PR applies pure RM but essentially on a local level, which allows it to be fast. It does this by dividing the model space into 8 sub-cubes, then dividing each of these again, and so forth, until the variation of the distribution function within a sub-cube falls to an acceptable level.

B: LIME<=1.6 achieves this by smoothing the points after the initial choice. There are two problems with this: it is an iterative process, and fairly slow; and it blurs out sharp boundaries in the point density. The present PR chooses points, not randomly, but sub-randomly, using a Halton-sequence algorithm, producing points which 'look' random but which avoid approaching too close to each other. Some careful attention was necessary to avoid artefacts at the sub-cube boundaries but test distributions of points from the total algorithm (8-tree recursive RM plus Halton-sequence choice of points in each sub-cube) ended up looking very convincing.

I added a macro TREE_POWER to lime.h whose value I have somewhat arbitrarily set at 2.1. The reason for this is as follows. LIME<=1.6 (at least with `par->sample==0 or ==2`) biases the point density by generating candidate radial distances `r` which have an `exp(-|k|r)` distribution. This, together with a 'flattened' RM test function which is the normalized bulk gas density raised to the power of 0.2, produces the broadly acceptable example-model point distributions of LIME<=1.6. When I started testing the present algorithm, I assumed that I could reproduce this by sampling the grid so as to follow the normalized bulk gas density, with neither exponential pre-bias nor flattening. I found this was not the case: the produced distribution is too sparse in the central portion of the example model and too dense in the outfield. The resulting central cells are clearly too big: rays traced through the centre exhibit far too much noise. Clearly the optical depth of the model is not a linear function of the density. I found I was able to reproduce the broad 'look-and-feel' of the <=1.6 grid distribution by raising the gas density function to the power in TREE_POWER. I rather regret now wiring this in as a macro, because it rather obscures the point that the user has now *complete freedom* to specify the grid density function in whatever way they please. Raising the gas density to a scalar power is a crude way to accomplish the desired end: there may be many more systematic ways to achieve this.

allegroLeiden:
Note that the Halton-sequence thing removes the need to smooth after grid point choice.

Issue #95, opened 2016-04-08
Raytracing is slow


christianbrinch:
Raytracing in version 1.5 seems incredibly slow (5-10 times slower) compared to previous versions. What happened?

ParfenovS:
The main loop in previous version of raytracing procedure was through grid points. The main loop in the new version is through pixels and within this loop there is also "anti-aliasing" loop.
In the case of example model the number of pixels 100x100 and the number grid points is 7000. Thus, for the new version one needs par->antialias*10000=40000 steps and the old version requires 7000 steps.

allegroLeiden:
I don't know what version you are looking at Sergey, because the loop was over pixels in 1.3.1, which was where I came in. 1.5 should be faster than that because all the sines and coses are now precalculated, calls to pow(blah,2) replaced by multiplications etc.

imckstewart:
I guess I ought to comment as myself rather than 'allegroLeiden'... one big change since 1.3 is that the procedure of following each ray was placed in a separate function. I guess that will add some overhead. It is better coding practice but if it turns out to slow things down massively I guess we ought to back it out again.

ParfenovS:
Yes, raytracing in version 1.3 is almost the same as in version 1.5. I meant version 1.4. The code of raytracing procedure of version 1.4 can be found in raytrace.c file (raytrace_1_4 function).

imckstewart:
Ok yeah, I remember that. I put in a PR a while ago (don't know why SM hasn't accepted it) for a raytrace which used some elements of that algorithm, but mainly I wanted to have better handling for pixels where the number of grid points per pixel was high, rather than the opposite extreme, which would be where any speed issue would be. So maybe this is something worth a look at.

tlunttil:
I had a quick look at where all that time spent in raytracing goes. It seems that most (more than half at least) of the time is spent in calls to the user-defined "velocity" function. This is with the default model.c, where the velocity function isn't even that complicated. So short of decreasing the number of rays traced (by reducing samples per pixel, which would worsen image quality), there's probably not much to be done, unless the whole velocity integration stuff is done differently.

allegroLeiden:
[...]

PR #96, opened 2016-04-11
Smoother raytracing output.

allegroLeiden:
The current raytracing algorithm leads to step-function artefacts in the channel maps it produces. These occur because the algorithm, although continuously sampling velocity at relatively finely-spaced intervals along the path of a ray, uses a single set of values for level populations throughout an entire Voronoi cell. In outer regions of an image where the Voronoi cells are likely to be significantly bigger than the image pixels, this leads to the step artefacts which can be seen in the image which is generated by the default example model file. The algorithm subject of the present PR removes these step artefacts by interpolating (to 1st order) the level populations within each Delaunay tetrahedron. This new algorithm has necessitated a lot of extra or altered code, from a new qhull() function (renamed to 'delaunay') which can return the necessary information about the tetrahedra, through routines to work out the sequence of cells passed through by any ray, to the new traceray_smooth() function to actually implement the algorithm.

To aid in checking these many changes I have divided the push into many smaller commits. I checked the intermediate code at several points with valgrind. The final code passes valgrind when the new algorithm is invoked, is backward compatible (because it defaults to the old), and with the old algorithm selected, still produces an identical cube to before.

The following points are worth noting:

  1. The awkward memory structure, particularly for grid points, has made some duplication of code hard to avoid; it also makes it difficult to free memory of grid attributes no longer required. I look forward to the time when we can implement a more flexible scheme.

  2. Initially the velocity was also interpolated, but this proved to be undesirable, because the variation in spectral line amplitude across a typical cell is decidedly non-linear, and a small error in velocity can lead to a large error in intensity. Therefore the present implementation of the routine samples the user-supplied velocity function at several places between entering and leaving a Delaunay tetrahedron.

  3. The interpolation is essentially a Finite Element scheme with 1st-order basis functions; an extension to 2nd order would be relatively easy to implement, requiring only level populations (plus dust/knu) to be stored also for points half-way between each pair of grid neighbours. A 2nd-order interpolation of velocity might also prove to be an adequate approximation for this quantity and thus allow the avoidance of calling a user-supplied function at the raytracing stage.

  4. The new algorithm is slower than the old, but there are many things which could be pre-calculated, if the ensuing cost in memory was thought to be worth it.


tlunttil:
Do you know how much slower approximately this new raytracing is compared to version 1.5?

christianbrinch:
No because the one time I tried 1.5, I gave up on the raytracing and killed it before it finished. The same model took 20 seconds to run in 1.4.



Sent from my iPhone

On 26. maj 2016, at 20.22, tlunttil <notifications@github.com<mailto:notifications@github.com>> wrote:


Do you know how much slower approximately this new raytracing is compared to version 1.5?

tlunttil:
> No because the one time I tried 1.5, I gave up on the raytracing and killed it before it finished. The same model took 20 seconds to run in 1.4.

Do you mean release version 1.5 or one with the changes in this PR included? I haven't tried this PR yet, but raytracing in release 1.5 shouldn't be that terribly slow (though definitely slower than in 1.4 if the image size is large).

christianbrinch:
The release version I suppose.

Sent from my iPhone

On 26. maj 2016, at 22.53, tlunttil <notifications@github.com<mailto:notifications@github.com>> wrote:


No because the one time I tried 1.5, I gave up on the raytracing and killed it before it finished. The same model took 20 seconds to run in 1.4.

Do you mean release version 1.5 or one with the changes in this PR included? I haven't tried this PR yet, but raytracing in release 1.5 shouldn't be that terribly slow (though definitely slower than in 1.4 if the image size is large).

tlunttil:
Hmm, that sounds strange. I wonder if it's a bug rather than a 'feature' in version 1.5 raytracing. I don't don't think a model that only takes 20 seconds in 1.4  should take very long even in 1.5. Can you send the model files so I could have a look?

allegroLeiden:
There seems to be a bit of a non-sequitur between @tlunttil 's first comment (which refers to the PR) and @christianbrinch 's reply to that (which refers to 1.5). The probable reason why 1.5 is slower has been remarked on already by @ParfenovS in thread #95.

On my desktop, for the master branch, which is still essentially 1.5, the time to run the standard model is about 13 min. Raytracing takes 25 sec of this. For branch 'raytrace_smooth' (the source of the present PR) with `par->traceRayAlgorithm=1` raytracing takes 120 sec. (Everything single-threaded.)

tlunttil:
Yeah, what I originally wanted to know is what you just told. Ok, so raytracing in this PR is a factor of 5-ish slower than release in 1.5. That doesn't sound too bad and it's still only a small fraction of the total run time.

I have some further comments and questions about 1.4 vs 1.5 raytracing, but it's better to continue that discussion in #95.

smaret:
Superseded by #145.

PR #100, opened 2016-04-11
Fixes #97 and #98

### LINKED ISSUE #97: X-axis is inverted in output images
Output images are flipped left-to-right. This was brought to our attention by @dquenard.

### LINKED ISSUE #98: Velocity inversion still for pregrid
The fix for #38 unfortunately was only partial: it neglected to fix the issue if the user chooses pregrid. Once again thanks to @dquenard for picking this up.

allegroLeiden:
. I flipped the sign of ray.x in line 267 of raytrace.c (plus some cosmetic cleanups
to 267 and 268): that should fix #97.
  . I flipped the sign of veloproject in line 148 of raytrace.c: that should fix #98.

### DIFF
--- a/src/raytrace.c
+++ b/src/raytrace.c
@@ -145,7 +145,7 @@ Note that the algorithm employed here is similar to that employed in the functio
 
               /* Calculate an approximate average line-shape function at deltav within the Voronoi cell. */
               if(!par->pregrid) velocityspline2(x,dx,ds,g[posn].mol[molI].binv,deltav,&vfac);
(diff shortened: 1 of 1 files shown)


allegroLeiden:
Note that, even if one sets DTEST in the makefile to freeze the random seed, that the output image from the default model.c will not be the exact left-right mirror inverse of the previous DTEST output, because the random dither value previously applied to pixel `img[im].pxls/2.0 + xi` is now applied to pixel `img[im].pxls/2.0 - xi` and so forth. This could be corrected but it does not seem worth the bother.

PR #102, opened 2016-04-12
Smallfixes 2

### LINKED ISSUE #74: Undefined 'FastExp' function warning
LIME gives the following warning during the compilation with gcc 5.2.0:

```
In file included from src/aux.c:10:0:
src/lime.h:250:15: warning: inline function 'FastExp' declared but never defined
 inline double FastExp(const float negarg);
```

allegroLeiden:
Fixes #74 and addresses #2.


smaret:
Thanks for the fix.

PR #71, opened 2015-09-03
Improved grid building

smaret:
This is PR contains @attilajuhasz's proposed changes for the grid building from PR #62.

PR #103, opened 2016-04-13
Added optional user function gridDensity

allegroLeiden:
This is intended to specify the probability distribution function for the internal (i.e.
non-sink) grid points. If the user does not supply their own version of this function,
it defaults to the algorithm which was previously used.

If this is accepted, I'll re-write the treegrid PR #80 (now #93) to incorporate it, which will avoid the necessity for a lot more user parameters.

This PR is relevant to issues discussed in #62 and will also overlap with or affect #71.


smaret:
Could you please document the `gridDensity` function in the the user manual?

smaret:
@attilajuhasz : Your PR #71 has a similar purpose than this one. Could you please have a look at what Ian proposed here and let us know what do you think?

smaret:
Looks good, but LIME seems to hang if `fracDensity` is set to values strictly lower than 1:

```c
void
gridDensity(inputPars par, double x, double y, double z, double *fracDensity){
  *fracDensity = .5; // this hangs
  /* fracDensity = 1.0; // this works */
}
```

smaret:
Also has `par->sampling `any effect if one sets `fracDensity` ?

allegroLeiden:
I'll look into these. Jeez you guys, it is either nothing or a shitload with you. ;)

smaret:
Thanks for looking into this.

> Add a warning to the documentation advising the user to choose a gridDensity() such that there is some place in the model where fracDensity>=1.

This sounds good to me. In fact I thought that `fracDensity` was some kind of probability (hence   comprised between 0 and 1) for a grid point to be chosen at a specific place. I think that a more detailed description of the function and its return value in the documentation would be useful.

allegroLeiden:
> In fact I thought that fracDensity was some kind of probability (hence comprised between 0 and 1) for a grid point to be chosen at a specific place.

That is exactly what I intended, and that is exactly the effect it *will* have, as soon as we can alter the other parts of the algorithm to become a more pure rejection method. Without doing the latter, there is no present way to achieve this.

smaret:
@allegroLeiden Can you please resolve the conflicts with the master branch?

PR #107, opened 2016-04-15
Fixes #106

### LINKED ISSUE #106: Wrong seeding of thread-private RNGs in levelPops()
In line 464 of aux.c the thread-private random number generators in levelPops() are seeded by the expression

    (int)gsl_rng_uniform(ran)*1e6

This is wrong, because the cast operates before the multiplication, giving a result which is always zero. Correct would be

    (int)(gsl_rng_uniform(ran)*1e6)

allegroLeiden:
This was a 1-line fix to #106. Note however that it will change the detailed values of the output image, even with DTEST set in the Makefile.

### DIFF
--- a/src/aux.c
+++ b/src/aux.c
@@ -461,7 +461,7 @@ levelPops(molData *m, inputPars *par, struct grid *g, int *popsdone){
 
   for (i=0;i<par->nThreads;i++){
     threadRans[i] = gsl_rng_alloc(ranNumGenType);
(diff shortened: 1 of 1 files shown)


tlunttil:
There looks to be the same bug in raytrace.c, line 210 of master branch. Not sure if you've already fixed it in some of your PRs.

imckstewart:
No, I hadn't noticed it. Well spotted (once again).

PR #108, opened 2016-04-21
Fixes #85

### LINKED ISSUE #85: Random crashes of LIME with the NO molecule
Dear all,

First of all, I would like to thank you again for your beautiful work on maintaining and developing LIME!

Please find in this link a folder containing all the files needed to reproduce our models:

[Click on me!](http://filez.irap.omp.eu/7c5p9hclbrzf)

We have tested several models to reproduce the emission of the NO molecule lines with LIME. To do that, we have defined a structure composed by a spherical source. We then generate a grid of points with this structure as an input for LIME. The structure parameters are the following:

Inner spherical source parameters

- Constant H2 density (1e6 cm^-3)

- Constant temperature (60 K)

- Constant abundance (1e-9)

- Size: 8000 AU

- Mass: 0.8 solar masses

- 2000 points grid (pregrid_lime.dat) attached to this report.

- Doppler parameter: 300

- Channel resolution : 250

- Bandwith (km/s) : 30

- Pixeld per dimension : 251

- Resolution (arcsec) : 0.2

- 1 NO line to model at this frequency 150.17648 (GHz)  

- NO collision file. We have modified this collision file obtained from the LAMDA database. We have deleted transitions with very low probabilities (Aij<1e-11 s^-1) just to speed up the computation since we have a lot of transitions in the original collision file. We have tested this new collision file in RADEX and is working fine.  We are attaching this file to this report as well. 

**We have noted that LIME crashes randomly several times when running this model. We have runned this very same model 10 times, and it crashes 9 out of 10 times**. We get this error message when it fails:

**gsl: svd.c:149: ERROR: SVD decomposition failed to converge
Default GSL error handler invoked.
/home/LimePackage/lime: line 92: 21362 Aborted                 ./lime_$$.x**

We are attaching the model.c file as well with the parameters we used for this particular test.

In addition, we have noted that the crashing time can vary significantly from one test to another when running the very same model with LIME several times (from 30 minutes to 8 hours).

**Another concern we have is, can we trust the results when one of the models is working, even when other tests with the same parameters have failed?**

Thank you very much for your support!

Cheers,

David Quénard
PhD student - Astrochemistry
Institut de Recherche en Astrophysique et Planétologie (Toulouse - France)

allegroLeiden:
The changes here are as follows:

- Added a warning if the density is too low at any grid point.
- LTE is now broken into 2 parts: the new function calculates the LTE solution for a single grid point and species.
- In stateq():
  - The SVD alternative has been removed.
  - The test of determinant==0 of the LU-decomposed matrix as a way to determine if the matrix is singular has been replaced by a test of the returned status value of the GSL LU solver.
  - The GSL error handler has been turned off for the duration of the matrix solution iteration loop so that failure and exit can be avoided in favour of an alternative way to estimate the populations at failing grid points.
  - LTE solution has been implemented as this alternative.
  - A warning message has been introduced if the LU fails.


smaret:
Duplicates #119.

PR #109, opened 2016-04-21
Faster raytracing

allegroLeiden:
@tlunttil made a suggestion in the discussion of #95 that the calls to velocity() inside traceray() could be moved outside the loops over channel and spectral line. This PR implements that change. The resulting speedup is more than a factor of 2.


tlunttil:
You could add schedule(dynamic) to #pragma omp for on line 295 of raytrace.c (before the loop over image pixels in raytrace() ). With static scheduling some threads finish significantly faster than others so come cpu cores will end up idling. In my tests with  8 threads dynamic scheduling made raytracing 20 to 30 percent faster.

smaret:
Superseded by #145.

Issue #111, opened 2016-04-22
Bad projection of grid point velocity.


allegroLeiden:
In line 211 of photon.c, the velocity component of the working grid point in the direction of the photon is calculated. However the grid point velocity is projected not onto the direction of the photon (initdir) but onto the direction of the first edge to be followed.

PR #112, opened 2016-04-22
Fix #110 and #111

### LINKED ISSUE #110: Bad test for sink points in photon().
The main loop which follows a photon through the model in function photon() exits at its conclusion if the grid point g[there] is a sink point. However, this really ought to be a test on g[here]. The reason for this is that new 'here' and 'there' values are chosen just prior to this end-of-loop test: the old 'there' becomes 'here' and a new 'there' is chosen depending on the direction chosen by sortangles(). This leaves the possibility open that the new 'here' for the next block pass could be a sink point. This is undesirable, although probably it does not cause any extremely bad effect.

allegroLeiden:
I also did a little simplification of this loop in photon.c. A proper job will have to wait for #83 to be accepted however.


smaret:
>  A proper job will have to wait for #83 to be accepted however.

I'm confused. Quoting @tlunttil in #125:

>  I think it's better to scrap #83 and rewrite the fixes.

tlunttil:
#83 was written before @allegroLeiden's much more extensive changes and merging them would probably require some (not insignificant amount of) work. I think you should ignore PR #83; I (or someone else?) will write a new PR with the same functionality in a way that fits better with Ian's PRs.

Issue #68, opened 2015-08-24
Angles in continuum polarisation source function


attilajuhasz:
I have a bit of a trouble understanding how the angles gamma and psi are calculated in the polarisation part of the continuum source function. My understanding is that gamma is the angle between the magnetic field (B) and the plane of the sky/image, while psi is the position angle of the magnetic field projected into the image plane (i.e. the angle between the direction of North in the image plane and the projected B-vector). The `sourceFunc_pol()` and `stokesangles()` functions consistently use a variable name `incl` to describe the orientation of the model/source. The essence of my question is: how are gamma and psi calculated in the polarisation module and what is the meaning of `incl` in the source function? From the usage of `incl` within the functions I'm guessing it should be the inclination angle. However in the raytracer it is `image.theta` that is passed to the polarisation source function and `image.theta` is equivalent of the inclination if and only if `image.phi` is zero. Also why is there only a single angle in the polarisation module to describe the source orientation as, to my understanding, psi depends on the position angle of the source, i.e. to calculate both gamma and psi the source orientation has to be described by two angles.

allegroLeiden:
I'm looking at this and I agree that the angle specification sent to stokesangles() is inadequate. I'm also wondering why the polarization fraction is hard-wired to 0.15. Going to ask Michiel about it.

PR #117, opened 2016-05-10
Fix for #68

allegroLeiden:
Fix for #68


tlunttil:
This PR looks good to me, but I have one suggestion/question:
Why is the check for polarization images done with an ugly and possibly error-prone test for a non-zero B field strength at a single point instead of (well, more like in addition to) par->polarization? The manual tells to set par->polarization if Stokes I/Q/U images are needed - there's nothing to indicate that setting B field alone is enough (and IMO, it shouldn't be). There's even a comment in the code saying that checking B!=0 isn't a good test, but is there some deep reason why it is still there?

A couple of very small cosmetic changes. I'm not sure if this PR is the right place to change these, or maybe these have already been done somewhere. They just caught my eye when I was reading the code.

- There's a magic-number AU-to-parsec conversion on line 170 of aux.c - it would be cleaner if such things were done with constants defined in lime.h.

- Function invSqrt (strarting line 338 of aux.c) isn't used (was it ever?) and should be removed. Anyway, the way it's done is not even a good idea on modern computers.

allegroLeiden:
This is why we need #77! I agree with you entirely @tlunttil about the test for B, which I have here only made marginally less evil than it was originally, but I had believed that `par->polarization` was an 'unsettable inpar' because whatever value the user might give it becomes overwritten at aux.c:141 or 155 (line nos in the PR version of aux.c). Apparently the doco believes otherwise.

I'm happy to change the code accordingly. I would (i) make `par->polarization` the sole means of controlling the polarization; (ii) still test the B field but only issue a warning `if(par->polarization && normBSquared<=0)`.

I agree about the other 2 points but this is not the only place in lime with a hard-wired value. I had in mind at some point (not necessarily pre-1.6) going through systematically and fixing them all. If it makes you happy though I will fix them here. :smile:

tlunttil:
I think the cosmetic changes are better left to another PR (and yes, probably post-1.6). It's better to do a general code clean-up all at once instead of piecemeal in PRs that are meant for something else. I guess we'd better make new issue for that.

tlunttil:
I don't think you need to (and you shouldn't) remove setting the default values at the beginning of parseInput. In your new commit you don't seem to call anywhere input (defined in model.c), where all the parameters are usually set.

Also: if (normBSquared > 0.) on line 61??

allegroLeiden:
@tlunttil the goal posts have moved considerably since Reinhold Schaaf's PR was merged in. Pseudo-code is now

    main.c:main(){
      main.c:initParImg(){
        // sets inpars defaults
        // reads them from input()
      }

      main.c:run(){
        aux.c:parseInput(){
          // does the remaining stuff with the pars
        }
        // does the remaining stuff which used to be in main()
      }
      // etc
    }

> Also: if (normBSquared > 0.) on line 61??

Sorry, what's the issue?

allegroLeiden:
> Sorry, what's the issue?

I'm an idiot. Thanks.

tlunttil:
> @tlunttil the goal posts have moved considerably since Reinhold Schaaf's PR was merged in. 

Ok,  I'd completely missed that somehow. In that case this PR looks good to me now.

smaret:
>  In that case this PR looks good to me now.

Thanks. @allegroLeiden: could you please merge the master branch into this branch so I can merge your PR?

allegroLeiden:
Done.

Issue #99, opened 2016-04-11
Line blending doesn't work properly


tlunttil:
Line blending is supposed to be taken care of with code between lines 257 and 280 in photon.c (main branch). However, the code doesn't seem to work properly. The for-loop starting at line 260 is supposed to go through all blended lines, but because the number of blended lines is calculated incorrectly, the loop body is never run. Parameter matrix is a pointer to struct of type blend, so sizeof(matrix) is on most machines 8 bytes, while sizeof(blend) is (at least) 16 bytes.

The code inside the loop (that is currently never run) also looks a bit suspicious to me. I don't think that the changes due to blending to tau and other parameters are necessarily made to the correct lines.

allegroLeiden:
I had a look at this. The loop maximum is easy to fix, but I think too that there is something screwy about the test `if(matrix[jline].line1 == jline` etc. The comparison should surely be against `iline` not `jline`, and the sign of the addition to `delta` should change depending on whether `.line1` or `.line2` matches `iline`. Maybe something like

    if(matrix[jline].line1 == iline || matrix[jline].line2 == iline){
      if(matrix[jline].line1 == iline)
        blendDV = deltav + matrix[jline].deltav;
      else
        blendDV = deltav - matrix[jline].deltav;

      if(!par->doPregrid)
        calcLineAmpSpline(g,here,dir,g[id].mol[counta[jline]].binv,blendDV,&vblend);

etc.

allegroLeiden:
I note too that the incoming spectrum in line overlap regions is calculated twice, i.e. once for each line. I suppose that makes for the simplest code however. I also don't much like that the line separation at which blending begins to be considered is hard-wired at 10^4. Seems to me it should be some multiple of the maximum Doppler width.

tlunttil:
I agree about the fixed blending threshold. Properly it should be something that depends on the ratio of max velocity separation in the model and max doppler width.

tlunttil:
The blending part is IMO written confusingly. Surely blending should be symmetric, so if line _a_ is blended with line _b_, so is line _b_ blended with line _a_. So we could have a symmetric matrix of zeros and ones where one indicates a (possible) overlap and a corresponding skew-symmetric matrix of line centre velocity differences. To check for blendings with the current `iline`, just find the ones on the _i_th row of the blending matrix. In practice the blending matrix would probably be mostly zeros, so it should be stored "sparsely". Something like this:

`
for (jline=0;jline<num_blends[iline];jline++) {
[...]

Issue #101, opened 2016-04-11
External grid points can fail to be labelled as sink points


allegroLeiden:
This problem was recognized by @tlunttil.

The current code labels as sink points only those which it places on the sphere which encloses the model volume. However, the algorithm which follows photons along links (at least as Tuomas has amended it) assumes that any non-sink point will have a further link which makes an angle of 90deg or less to the photon direction of travel. However, it is possible to generate nominally internal points for which this is not true. Such points lie on the convex hull of the entire set of grid points, but at radii less than the sphere radius. They are really sink points, but do not get labelled as such. The likelihood of this occurring will tend to be greater when the number of sink points demanded is relatively small; this may be the original reason that the number of sink points in the example model file seems rather excessive.

In principle it is easy to identify such points, because the Delaunay tetrahedra of which they are vertices will have no neighbours on the external faces. Such faces can be identified with the aid of the same qhull routines that are used for the triangulation. The difficulty with making use of this information is that the present way in which the grid information is stored makes it difficult to adjust the number of sink points. This is yet another argument for a reorganization (see #81).

PR #118, opened 2016-05-10
Fix for #99

allegroLeiden:
Fix the line blending (#99)


ParfenovS:
nextLineWithBlend will be incremented on each cycle of matrix inversion loop (stateq.c file) which begins with:
while((diff>TOL && iter<MAXITER) || iter<5){
getjbar(id,m,g,ispec,par,blends,nextLineWithBlend,mp,halfFirstDs);

it is strange. Should one replace this by:

while((diff>TOL && iter<MAXITER) || iter<5){
*nextLineWithBlend = 0;
getjbar(id,m,g,ispec,par,blends,nextLineWithBlend,mp,halfFirstDs);

ParfenovS:
Or maybe one should use some auxiliary variable. For example

while((diff>TOL && iter<MAXITER) || iter<5){
int nextLineWithBlend1 = *nextLineWithBlend;
getjbar(id,m,g,ispec,par,blends,&nextLineWithBlend1,mp,halfFirstDs);

and after the end of matrix inversion loop in stateq one can add the following line:

*nextLineWithBlend = nextLineWithBlend1;

tlunttil:
I agree with @ParfenovS , the second version with the auxiliary variable should work. Setting *nextLineWithBlend to zero would probably be ok unless there are several species.

ParfenovS:
I think that there is also an error in raytracing.
If the user specifies img.trans then the code assumes that this transition belongs to the first species and 
`img[im].freq=m[0].freq[img[im].trans];`  (see raytrace procedure)

This PR contains following code in the traceray procedure:

`if(img[im].trans > -1){
                    lineRedShift=(m[molI].freq[img[im].trans]-m[molI].freq[lineI])/m[molI].freq[img[im].trans]*CLIGHT;`

where molI changes in the loop. This part should be replaced by 

`if(img[im].trans > -1){
                    lineRedShift=(m[0].freq[img[im].trans]-m[molI].freq[lineI])/m[0].freq[img[im].trans]*CLIGHT;`

ParfenovS:
I just noticed that the master branch also contains this error.

ParfenovS:
Actually, whole this part:
`if(img[im].trans > -1){
  shift=(m[0].freq[img[im].trans]-m[molI].freq[lineI])/m[0].freq[img[im].trans]*CLIGHT;
} else {
  shift=(img[im].freq-m[molI].freq[lineI])/img[im].freq*CLIGHT;
}`

can be safely replaced by a single line:
`shift=(img[im].freq-m[molI].freq[lineI])/img[im].freq*CLIGHT;`

tlunttil:
Right, there's no need for `if(img[im].trans > -1)` part, because `img[im].freq` is set in `raytrace()` (line 220 in master branch).

This bug, too, only seems to have an effect if there are multiple species. I'm not surprised if there are some further bugs with multi-species calculations, those parts have probably not been used/tested much.

allegroLeiden:
Thanks for pointing this out @ParfenovS. Damn - I thought this `nextLineWithBlend` thing was rather a clever idea, but it does depend on the `molI` and `lineI` loops always being nested with no other layers of loop in between, which as you point out is alas not the case in the sequence of calls which ends in `getjbar()`.

allegroLeiden:
I have fixed the `getjbar()` problem now. I decided to do this via a redesign of the `blendInfo` struct. Pseudocode for the whole call nesting from within 'levelPops()' to `getjbar()` is as follows:

    nextMolWithBlend = 0;
    for(ispec=0;ispec<par->nSpecies;ispec++){
      stateq(ispec, nextMolWithBlend, etc){
        while((diff>TOL && iter<MAXITER) || iter<5){
          getjbar(ispec, nextMolWithBlend, etc){
            for(iphot=0;iphot<g[posn].nphot;iphot++){
              if(mp[molI].vfac[iphot]>0){

                nextLineWithBlend = 0;
                for(lineI=0;lineI<m[molI].nline;lineI++){
                  // blah
                  if(par->blend && molI==blends.mols[nextMolWithBlend].molI\
                  && lineI==blends.mols[nextMolWithBlend].lines[nextLineWithBlend].lineI){
                    // do blend stuff

                    nextLineWithBlend++;
                    if(nextLineWithBlend>=blends.mols[nextMolWithBlend].numLinesWithBlends)
                      nextLineWithBlend = 0; // A dummy value. If we get here, there are no more blends, but we don't want the matrix calls which will blindly check that to read memory out of range. Could have done this cleverer but this works.
                  } // end if mol.line has blends
                } // end loop lineI

              } // end if
            } // end loop iphot
          } // end getjbar()
        } // end while
      } // end stateq()

      if(this molecule had some lines with blends)
        nextMolWithBlend++;
    } // end loop ispec

allegroLeiden:
I'll look at the img.trans thing now, but honestly, sometimes I despair at the task of (i) detecting such problems and (ii) acceptably patching them in ways which don't add more ugliness. This whole style of guess-my-mode-from-some-possibly-irrelevant-values really needs to be replaced by a proper set of booleans. Separating the user params from internal configuration information via PR #77 would be a good launching pad for this, but SM hasn't yet accepted it...

ParfenovS:
The new version of this PR causes segmentation fault in the case of blended lines absence.
The lineBlend procedure frees blends structure if there are no blending lines. In the result, for example, the following statement (photon.c file in this PR) will rise segmentation fault:
`if(par->blend && molI==blends.mols[nextMolWithBlend].molI\
          && lineI==blends.mols[nextMolWithBlend].lines[nextLineWithBlend].lineI){`
because it tries to access deallocated blends.mols array.

Possible fix may look like:
`if(par->blend && blends.mols!=NULL){`
          `if(molI==blends.mols[nextMolWithBlend].molI\
          && lineI==blends.mols[nextMolWithBlend].lines[nextLineWithBlend].lineI){`

tlunttil:
It can be done within a single if because C short-circuits evaluation of &&:
`if((par->blend) && (par->blends.mols) && (molI==blends.mols[nextMolWithBlend].mol)I\ && (lineI==blends.mols[nextMolWithBlend].lines[nextLineWithBlend].lineI)) {`

You also need to make sure that `par->blend.mols` gets set to NULL after freeing, it doesn't seem to be done at least in 'freeBlends'.

allegroLeiden:
Thanks guys!

allegroLeiden:
Tuomas I guess your 'par->blends.mols' is a typo. I set blends.mols to NULL after freeing it within lineBlend() but not after the free at the end of levelPops. I guess it might be more conservative practice to set things to NULL inside the various free*() functions, but the pattern in Lime is not to do that, so I think any change should be post-merge, or at least not piecemeal.

tlunttil:
Yes, I meant `blends.mols` inside that `if`. And somehow I failed to see that while you don't set the pointer to NULL inside `freeBlends`, you do it in the next line after calling that...

In general it would probably be cleaner if the various `free*()`s did set pointers to NULL after freeing, but I agree it's more of a nice-to-have at this point, there are more important things to do first.

tlunttil:
There are a couple of strange things (from time immemorial, not new in this PR, but they're sort of related to all this).

-  Line 254 of photon.c (in this PR): why vfac[0] instead of vfac[molI]?
-  In fact, why is the local vfac (that is stored in mp[molI].vfac[iphot]) done the way it is in the first place. Wouldn't it be better if it were strictly local without including the half-step thingy? So it would essentially be just an evaluation of a Gaussian at deltav.

New getNextEdge will probably make some (currently working) runs fail due to 'no backwards edges' check ), but I don't know if that is a problem. Anyway, the fundamental problem #101 is in the sink point placement/usage.

allegroLeiden:
- vfac[0] is probably a typo. These merges are pretty tricky.
- I have no strong opinion but I would guess this belongs with the other fixes to photon propagation which I hope you will pr some time.

tlunttil:
Right, I'll think about the vfac thing and maybe do that together with the other propagation stuff.

My only question is then whether the bailout in case of 'no edges in the forward half-space' is the right way to go? Current version would happily(?) run in such cases, which can easily happen if the number of boundary sink points is lowish. The whole sink issue needs to be dealt with at some point, but a good solution won't be ready for 1.6. I'm wondering if instead of bailout it would be better to only issue a warning and choose the 'least-bad' edge.

allegroLeiden:
> I'm wondering if instead of bailout it would be better to only issue a warning and choose the 'least-bad' edge.

Yeah, ok. This is probably a better idea.

By the way, I changed vfac[0] to vfac[molI] already.

allegroLeiden:
> Right, I'll think about the vfac thing and maybe do that together with the other propagation stuff.

Or by 'vfac thing' do you mean 'why do vfac for the first step in the first place'. I guess you do.

tlunttil:
> Or by 'vfac thing' do you mean 'why do vfac for the first step in the first place'. I guess you do.

Yes ;)  

Re: edge selection, allowing backtracking can of course lead to infinite loops, which is how I first found the problem. But if the user is given a warning, at least getting stuck won't come as a complete surprise...

tlunttil:
I think this one's ready for merging, too. And thanks to @ParfenovS for spotting those problems.

allegroLeiden:
Time to resolve the conflicts @smaret ?

smaret:
@allegroLeiden Yes, please do.

PR #119, opened 2016-05-10
Fix for #85

allegroLeiden:
The changes here are as follows:

- Added a warning if the density is too low at any grid point.
- LTE is now broken into 2 parts: the new function calculates the LTE solution for a single grid point and species.
- In stateq():
  - The SVD alternative has been removed.
  - The test of determinant==0 of the LU-decomposed matrix as a way to determine if the matrix is singular has been replaced by a test of the returned status value of the GSL LU solver.
  - The GSL error handler has been turned off for the duration of the matrix solution iteration loop so that failure and exit can be avoided in favour of an alternative way to estimate the populations at failing grid points.
  - LTE solution has been implemented as this alternative.
  - A warning message has been introduced if the LU fails.

### DIFF
--- a/src/LTEsolution.c
+++ b/src/LTEsolution.c
@@ -28,7 +28,7 @@ void lteOnePoint(inputPars *par, molData *m, const int ispec, const double temp,
 
   sum = 0.0;
   for(ilev=0;ilev<m[ispec].nlev;ilev++){
-    pops[ilev] = m[ispec].gstat[ilev]*exp(-100*CLIGHT*HPLANCK*m[ispec].eterm[ilev]/(KBOLTZ*temp));
+    pops[ilev] = m[ispec].gstat[ilev]*exp(-HCKB*m[ispec].eterm[ilev]/temp);
     sum += pops[ilev];
   }
   for(ilev=0;ilev<m[ispec].nlev;ilev++)
--- a/src/LTEsolution.c
+++ b/src/LTEsolution.c
@@ -11,19 +11,12 @@
 
 void
 LTE(inputPars *par, struct grid *g, molData *m){
-  int id,ilev,ispec;
-  double z;
+  int id,ispec;
 
-  for(ispec=0;ispec<par->nSpecies;ispec++){
-    for(id=0;id<par->pIntensity;id++){
+  for(id=0;id<par->pIntensity;id++){
+    for(ispec=0;ispec<par->nSpecies;ispec++){
       g[id].nmol[ispec]=g[id].abun[ispec]*g[id].dens[0];
-      z=0;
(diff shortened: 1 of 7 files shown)


allegroLeiden:
I just noticed that this is essentially the same PR as #108! I'm not sure how that happened. #108 should be rejected; this one is slightly better.

tlunttil:
I'm not sure about the gsl error handling part: according to [https://www.gnu.org/software/gsl/manual/html_node/Error-Handlers.html#Error-Handlers](url) function `gsl_set_error_handler()` shouldn't be used in multi-threaded programs except for setting a program-wide handler in the master thread. On line 101 of stateq.c you however use that function to reset the default handler (inside a multithreaded part). Is it even necessary to do that anyway? Why not just turn off the handler at the start of the program and be done with it? That's what the gsl documentation recommends for production programs, too.

allegroLeiden:
Could be you are right... I was just following conservative instincts. It seemed to me better to trap errors in a controlled way than to leave the program open to seg faults. Surely, if one switches off the default handler at the start, doesn't that mean that one ought to write bespoke handlers?

Anyway I'll have a more informed opinion after I have read the doco a bit more thoroughly...

tlunttil:
You mean checking the return value of every call to gsl functions? I guess, sort of... But most uses of gsl are just calls to RNGs, which don't seem to report errors anyway. Otherwise it's just linear solvers in stateq, which you've already taken care of, and various bits and pieces during initialisation (calculation of polynomial coefficients in velospline etc.) so it's not a huge number of places. 

Anyway, to quote gsl documentation part 3.1:

> The purpose of the gsl_error handler is to provide a function where a breakpoint can be set that will  catch library errors when running under the debugger. It is not intended for use in production programs, which should handle any errors using the return codes.

allegroLeiden:
@tlunttil I think for the present I am just going to switch off the default handler before the parallel section(s). There are quite a few gsl calls in Lime (grep lists 2 pages of them) and the real issue, it seems to me, is that most of them have no error handling at all. Until somebody gets around to writing some ( @aschmiedeke ? :smile: ), it seems to me better to have at least something to catch errors, no matter how unlikely they may be.

tlunttil:
If guess about half of gsl calls are to RNGs for which there's no need (and apparently no possibility) for error handling. The rest is mostly in the initialisation phase where I agree some error handling is needed.

tlunttil:
This PR looks good to me. To prove that I really did read the code, I'll give one suggestion :wink: 
In lteOnePoint you could use constant HCKB (defined in lime.h) to clean up the argument of the exponential function.

smaret:
Thanks @imckstewart for the PR and @tlunttil for reviewing it!

Issue #127, opened 2016-05-18
Potential collPart mismatch in getmatrix()


imckstewart:
The function assumes that the number of density values is given by `m[ispec].npart`. This is unsafe and unrobust.

Issue #126, opened 2016-05-18
Redundant calculation of knu, dust for `par.restart && !img.doline`


imckstewart:
`par.restart` causes `popsin()` to be called, which reads values of `knu` and `dust` for each grid point. A non-line image however calls `continuumSetup()` which calls `kappa()` which resets the values of `knu` and `dust` - or at least attempts to, since the calculation of `knu` requires `dens` which is not loaded in `popsin()` or in any other function which is called in this sequence. The result is a seg fault.

Issue #123, opened 2016-05-12
Mol density treated differently for LTE and non-LTE


allegroLeiden:
For non-LTE there is some attempt made to tailor the calculation of the `nmol` values per grid cell per species to the collision partners supplied, but for LTE `nmol` is simply calculated proportional to the 1st density element.

Issue #122, opened 2016-05-12
m[i].lcl, lcu malloc'd to the number of transitions of the 1st coll partner.


allegroLeiden:
See the 4 lines in molint.c starting at 201. There should be separate `lcl`, `lcu` for each collision partner.

PR #128, opened 2016-05-20
Allows better user control of density/dust calculations

allegroLeiden:
This PR fixes or addresses 8 issues: #58, #86, #91, #121, #122, #123, #126 and #127. The central change is the introduction of new (optional) user parameters `collPartIds`, `nMolWeights` and `dustWeights`. These are all in effect vectors and, if set at all, are each expected to have the same number of valid elements as are returned by the user's `density()` function. They are made optional to preserve backward compatibility; if any or all are left unset, the task will attempt to reproduce the old behaviour.

To allow consistent and systematic use of these parameters it was necessary to cut the functions `molinit()` and `kappa()` into smaller units which parcel out the functionality in more logical units. This rearrangement also provided opportunities to fix some other tangles and inconsistencies.


tlunttil:
Now that you've moved all `free()`s to one place, maybe you could consider removing the unnecessary checks for `NULL` (as said in #73). Functionally it makes no difference, but I find the numerous lines of `if (ptr!=NULL)` annoying.  Then you'd clear 9 issues with the same PR...

allegroLeiden:
:smile: I wondered if you would remark on that. Ok - I'll think about this a bit more, but I'll probably do it, since you ask so nicely.

Some time the team should probably decide on a 'house style' for Lime. It is a bit of a mixture at the moment, which I think gives contributors more license to pull the code in their own stylistic direction, which I don't think is desirable.

tlunttil:
Speaking of style: I would very much prefer if variables weren't declared only at the beginning of each function. Even with ANSI C the variables only needed inside a scope block (a loop for example) could (and should) be declared inside the scope block. With c99 standard (which all compilers support nowadays) even loop variables could be done like `for (int i=0;i<10;i++)`. This would make the code cleaner and less error prone. For example, there'd be less chance of error in not declaring a variable private for an parallel (OMP) region, because locally declared variables are automatically private. 

Not a hugely important thing, and I'm not really looking forward to changing this in every file...

smaret:
Could please document these new functions in the user's manuals? Perhaps it would be also useful to have an example model file in the distribution which several collisional partners in it (e.g. CO with o-H2 / p-H2) ?

doc/usermanual.rst
```diff
@@ -639,25 +664,14 @@ density profile of the collision partner(s).
...
-in the density subroutine than there are collision partners listed in
-the molecular data file. In most cases, a single density profile will
-suffice. The density is a number density, that is, the number of
-collision partners per unit volume (in cubic meters, not cubic
-centimeters). Please note that the current version of LIME always takes
-the abundance relative to the first collision partner. This is
-potentially a problem if the first collision partner is not the total H2
-density and the user will have to correct for this in the abundances
-(see below).
+LIME can deal with an unlimited number n of collision partners (CPs). In most cases, a single density profile will suffice. Note that the number of returned density function values no longer has to be the same as the number of CPs listed in the moldata file(s) so long as the user sets values for the collPartIds parameter, but if this parameter is not supplied, and the numbers are different, LIME may not be able to match the CPs associated with each density value to those in the moldata file(s). Note also that moldata CPs for which there is no matching density will be ignored.
```
smaret: > unlimited number n of collision partners

Isn't this limited by  `MAX_N_COLL_PART`?

smaret:
Great stuff!

I've tested your PR with this [model file](https://gist.github.com/smaret/9929d9b8565ee4f0c93086a2767bba03). This is essentially the same model as the example included in LIME distribution, but with CO instead of HCO(+). I used the [co.dat](http://home.strw.leidenuniv.nl/~moldata/datafiles/co.dat) collision coefficient file (which includes both o-H2 and p-H2 as collision partners) from the LAMBA database. Everything works as expected.

I've made a couple of minor changes in the example model file:

1. Comment lines are wrap to make the reading easier
2. I've set the new parameters explicitly since this what we recommend. This also avoid a warning message with the model.

My comment on the documentation is very minor so I'm going to merge this PR as it is. Please feel free to edit the documentation if needed.

allegroLeiden:
I should edit the documentation directly in lime-rt:master?

smaret:
Yes.

PR #133, opened 2016-06-03
Fixes #131 by setting grid.nmol for sink points when par.lte_only.

### LINKED ISSUE #131: .nmol never set for sink points when `par.lte_only`.
Inside the raytracing algorithm, i.e. the block starting at line 111 of raytrace.c, `posn` can often be a sink point. Thus `sourceFunc_line()` and `sourceFunc_cont()` are called for sink points. These functions require values to be set of the grid-point attributes `nmol`, `binv`, `*pops`, `*knu`, `*dust`. Of these, all are given some value for all points for non-LTE, but `nmol` is not set for sink points for LTE.

allegroLeiden:


smaret:
This fix is also included in #132, so I'm closing this one as duplicate.

PR #134, opened 2016-06-09
Allows easy integration of LIME into ARTIST

rschaaf-aifa:
One central element of ARTIST is a SWIG interface to LIME to allow calling LIME from Python. Since LIME is designed as a stand-alone program, integrating LIME into ARTIST so far requires changes in LIME's code, mainly in aux.c where the input parameters and images are allocated and initialized.

In this PR, the critical parts of `parseInput` and `freeInput` in aux.c are moved to the new functions `initParImg` and `freeParImg` in main.c. In addition, the `main` function is split into the functions `run` and `main` (both in main.c), where `run` can serve as an interface to LIME for external programs (such as ARTIST), and `main` is used for stand-alone LIME.

Besides this structural change, some minor changes are necessary for easy integration of LIME into ARTIST:
- Made `silent` a global variable instead of a global constant
- Added a standard `#ifndef #define` construct to lime.h to avoid problems with including it more than once
- Set locale to "C" explicitely to avoid problems when reading in files
- Removed if statement around `par->nthreads = NTHREADS`

As a result of these changes, LIME's code can remain unchanged when integrating it into ARTIST which will make integration of future LIME versions much easier than in the past.

### DIFF
--- a/src/aux.c
+++ b/src/aux.c
@@ -19,17 +19,6 @@ parseInput(inputPars *par, image **img, molData **m){
   double BB[3];
   double cosPhi,sinPhi,cosTheta,sinTheta,dummyVel[DIM];
 
-  /* Allocate moldata array */
-  if(par->nSpecies > 1) {
-    (*m)=malloc(sizeof(molData)*par->nSpecies);
-    for(ispec=0;ispec<par->nSpecies;ispec++)
-      m[ispec]->nlev=0;
-  }
-  else {
-    (*m)=malloc(sizeof(molData)*1);
-    m[0]->nlev=0;
-  }
-
   par->ncell=par->pIntensity+par->sinkPoints;
   par->radiusSqu=par->radius*par->radius;
   par->minScaleSqu=par->minScale*par->minScale;
--- a/src/aux.c
+++ b/src/aux.c
@@ -57,16 +57,6 @@ The cutoff will be the value of abs(x) for which the error in the exact expressi
   */
   par->taylorCutoff = pow(24.*DBL_EPSILON, 0.25);
 
-  if(par->dust != NULL){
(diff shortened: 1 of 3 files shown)


allegroLeiden:
Otherwise the PR looks fine.

rschaaf-aifa:
Thank you @allegroLeiden for your comments. 

ad 1) You are right. I removed the now unnecessary check from aux.c.

ad 2) Again you are right what the malloc of `molData` is concerned; I removed the new code as you suggested. I am not sure what exactly you mean with the test of `par->nSpecies`, I could not find these lines in the code.

ad 3) These functions are used only within main.c, and will most probably never be used elsewhere. So there is no need to add them in the (already too long) `lime.h`. However, to make this clearer, I added appropriate forward declarations in `main.c`.

ad 4) OK

ad 5) Fair point. I am using several editors/IDEs on more than one computer, so setting all these tools up properly is a bit of a nuisance. I will try to improve on this point.

smaret:
@rschaaf-aifa Did you close this PR intentionally?

rschaaf-aifa:
Oops, sorry, that was a mistake. I am still not familiar enough with github...

allegroLeiden:
@rschaaf-aifa re the test of `par->nSpecies`: the first example (which began at an `else` statement at line 29 of aux.c in your previous code, and part of the block mallocing the molData object) has now been removed; the second example, namely the line

    if( par->nSpecies == 0 )

was formerly at line 79 of main.c and is now at line 87. As I remarked previously, the `==` here should probably be changed to `<=`, but this is now independent of your PR and were therefore better to be done elsewhere.

rschaaf-aifa:
Although independent of the main purpose of the PR, I followed your suggestion @allegroLeiden and changed the test accordingly.

Issue #81, opened 2015-09-25
Rearrangement of struct grid


imckstewart:
Github stuffs around with my fixed-space formatting, so my apologies that this looks so messy. I tried to attach it as a text file, but it was not accepted.
vvvvv
I'd like to instigate a discussion on restructuring of the 'grid' struct. There are two reasons for doing so: (i) to allow precalculation of a 'vfac spectrum' for each link or edge connecting two grid points, which could speed up processing dramatically; (ii) to improve the present somewhat disorganized memory layout.

Most of the data which is stored in 'grid' falls naturally into several arrays of varying dimension and size as follows:

  Type      Name              Dimensions
--------------------------------------------------------------
  int       id                par->ncell
  double    x                 par->ncell * DIM
  double    vel               par->ncell * DIM
  int       numNeigh          par->ncell
  int       sink              par->ncell
  int       nphot             par->ncell
  int       conv              par->ncell
  double    dopb              par->ncell
  double    dens              par->ncell * par->collPart
  double    t                 par->ncell * 2
  double    nmol              par->ncell * par->nSpecies
  double    abun              par->ncell * par->nSpecies
  double    mol.dopb          par->ncell * par->nSpecies
  double    mol.binv          par->ncell * par->nSpecies
--------------------------------------------------------------

The elements I have omitted are of two types: those whose size scales with the number of nearest neighbours of the point, and those whose size depends on the molecular species. A list of the first set is

------------------
  point       *dir;
  struct grid **neigh;
  double      *w; // doesn't seem to be used.
  double      *a0,*a1,*a2,*a3,*a4;
  double      *ds;
------------------

The species-dependent set is contained in the element 'mol' of type struct populations. Their sizes are tabulated as follows:

  Type      Name              Dimensions
--------------------------------------------------------------
  double    mol.pops          par->ncell * par->nSpecies * m[i].nlev
  double    mol.knu           par->ncell * par->nSpecies * m[i].nline
  double    mol.dust          par->ncell * par->nSpecies * m[i].nline
  double    mol.partner.up    par->ncell * par->nSpecies * m[i].npart * m[i].ntrans[j]
  double    mol.partner.down  par->ncell * par->nSpecies * m[i].npart * m[i].ntrans[j]
--------------------------------------------------------------

I discuss here better memory arrangements for the three sets of attributes in three respective sections.

A) Nearest-neighbour quantities.
================================
[...]

PR #135, opened 2016-06-09
On-the-fly interpolation of collision coeffs.

tlunttil:
Implementation of the on-the-fly interpolation of collisional
coefficients that was discussed in comments to issue #81. With the
default example/model.c memory saving is about 50% and there is no
effect on speed.

### DIFF
--- a/src/aux.c
+++ b/src/aux.c
@@ -215,8 +215,8 @@ The cutoff will be the value of abs(x) for which the error in the exact expressi
       (*m)[i].freq = NULL;
       (*m)[i].beinstu = NULL;
       (*m)[i].beinstl = NULL;
-      (*m)[i].up = NULL;
       (*m)[i].down = NULL;
+      (*m)[i].ntemp = NULL;
       (*m)[i].eterm = NULL;
       (*m)[i].gstat = NULL;
       (*m)[i].cmb = NULL;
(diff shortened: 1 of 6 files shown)


allegroLeiden:
This all looks good, I just have a few stylistic quibbles as follows:

- molinit.c, line 271: the loop could and should be omitted, together with the commented-out lines within (also other commented-out lines in molinit()).

- molinit.c, lines 290-295: this 'else' block (inherited from the previous code) could be made a small bit more conservative/safer by replacing it with something like

          } else if(g[id].t[0]<=part[ipart].temp[0]) {
            g[id].mol[i].partner[ipart].t_binlow=0;
            g[id].mol[i].partner[ipart].interp_coeff=0.0;
          } else {
            g[id].mol[i].partner[ipart].t_binlow=ntemp[ipart]-2;
            g[id].mol[i].partner[ipart].interp_coeff=1.0;
          }

- stateq.c, line 128 (applies also to line 116 as inherited from the previous code): personally I would not use 't' here, since it makes one think of temperature. 'ti' perhaps, with 'li' at line 116.

- stateq.c, lines 129-131: instead of 'trans_offset' defined as given, why not just something like

    `int i = t*m[ispec].ntemp[ipart]+g[id].mol[ispec].partner[ipart].t_binlow;`

Then the next 2 lines (actually a single line, wrapped) would become much more compact/readable as well as more conservatively expressed:

      double down = downrates[i] + g[id].mol[ispec].partner[ipart].interp_coeff*(downrates[i+1] - downrates[i]);

tlunttil:
Yep, I forgot to remove the commented-out old parts. I'll do that, and your other suggested changes, and push a new version probably later today.

PR #142, opened 2016-06-21
Fixed #137

### LINKED ISSUE #137: Too long before presence of user functions is tested.
In `buildGrid()`, the user functions `density()` etc are not accessed until after the smoothing, which can take a minute or so. The presence of such functions in `model.c` is not tested until that point. It would be 'nicer' if the presence of the 5 mandatory functions was tested before smoothing so the user does not have to wait to be notified if they have failed to include one of them.

allegroLeiden:
The presence of the 5 mandatory functions (mandatory under certain circumstances that is) is checked now early in the processing so the user does not have to wait to be told if they have made this kind of elementary error.

I also fixed 2 singularities in model.c and cleaned up its formatting.

### DIFF
--- a/example/model.c
+++ b/example/model.c
@@ -13,35 +13,38 @@
 
 void
 input(inputPars *par, image *img){
+  int i;
+
   /*
    * Basic parameters. See cheat sheet for details.
    */
-  par->radius			= 2000*AU;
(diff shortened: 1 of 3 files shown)


smaret:
Looks good, thanks!

Issue #21, opened 2014-12-04
Implement multi species support


mwestphal:
@christianbrinch  Around lime source, there are multiple instance of use of m[0] and abun[0]
If i understand well they point respectively to molData of first specie and abundance of first specie.

Is the first specie  more special than the others ?

tlunttil:
In addition to taking care of line blending due to multiple species in MC ray tracing calculations (see #99), `stateq()` needs to be updated to solve the level populations of all cross-excited species simultaneously. It seems to me that this would need quite a lot of work... Is multiple species with cross-excitation an important feature that people need?

PR #145, opened 2016-06-29
Combined raytrace

allegroLeiden:
This is intended to be the final word (at least in the framework of enhancements discussed for 1.6) on raytrace - or my final word anyway. Here I have merged and thus superseded the code which is in PRs #79, #96 and #109, also I have added new code which is a modified version of the fast raytrace algorithm which was in LIME 1.4 but which was (for various reasons) not propagated to 1.5. This PR thus answers the issue raised by @christianbrinch in #95 .


smaret:
@christianbrinch, @tlunttil : can you help reviewing this one? You know the raytracing algorithm much better than I.

tlunttil:
I can help with this review.

tlunttil:
Ok, this is a big one, but luckily it's written clearly and it even has useful comments. I can't claim that I went through the code thinking about every possible corner case of rays passing through the Delaunay tetrahedra in weird ways, but in general this looks good to me. Some notes:

- This PR (and maybe also some others) uses _Bool. I'd include stdbool.h and use bool instead. Anyways, these are C99 additions - I'm not sure if some older gcc version that someone could still conceivably use would default to an older standard.
- freePop2 has unnecessary checks for NULL before free()s,
- There looks to be a lot of room for optimisation in the new raytracer, mainly by precalcing various things (as is also noted in a comment in the file) and some loop restructuring. But I think that performance tuning is better left for a later time.
- According to the qhull [documentation,](http://www.qhull.org/html/qdelaun.htm) by default co-spherical regions are merged so that 'facets' are not necessarily simplical (i.e. tetrahedra in 3D). I'm not sure if such a case would break @allegroLeiden's code. Maybe qhull should be forced to make simplical facets?
- There are various places where the factor 100. * HPLANCK * CLIGHT/KBOLTZ  could be replaced with HCKB. It might be worth greping for this in all files.

- (General comment, not only related to this PR: there seems to be a lot of stuff in lime.h that wouldn't need to be there, for instance function prototypes for functions that are only used in a single .c file. (Actually making them visible to everyone may even make automatic inlining impossible?))
- (Also, inside functions I would prefer moving variable definitions from all-at-the-beginning to whatever scope they're used in. And also maybe using C99 style loop variables: for (int i=0;...).)
- (Unrelated to this PR, I only noticed because it's adjacent to one of the changes in the makefile: functions in sourcefile getclosest.c aren't used for anything so the whole file could be removed.)
- (Another not-this-PR thing related to the makefile: there's a lot of warnings (with gcc at least) about unused results due to not checking the return value of fread()s etc. Should -Wno-unused-result be added to compiler directives to suppress the warnings?)

allegroLeiden:
- C99: we run some machines which have RHEL6 on them for purposes of running CASA. The gcc version on these (4.4.7) complains about c99 stuff, desires an option -std=c99 or -std=gnu99. I guess we could do that, and I don't have a really strong opinion here, but I would lean a bit in the direction against adopting c99 practices, unless there is a clear need. Comments @smaret @ParfenovS ?
- I'll get rid of the NULL checks. Ditto ->HCKB.
- qhull: the sooner we can move away from this pig the happier I'll be. I've always found the documentation extraordinarily difficult to fathom and I am always a bit surprised when I succeed in getting it to work the way I want. By all means, if there is a way to force simplicial facets then we should adopt this.
- Re the general comments: I suggest we open an issue for all the cleanups we want to undertake - that way we can add or debate stuff in common. Either that or start compiling a list of things on the wiki. Or both: issue and wiki.

tlunttil:
Yeah, calling qhull from C isn't the easiest thing, which is what even the documentation says...  Unfortunately I don't know a good (well, more like any) alternative for qhull that can be called from C. 

It looks like option Qt should work for forcing simplical facets. In practice I think it's usually only a few facets that are not simplical.

For the general comments, including what standard of C to use, I guess a new issue would be a good place to discuss that. But I'll probably wait until after we're done with 1.6.

allegroLeiden:
I would love to write a triangulator myself, it would be great fun. However I don't think I can justify spending the time on it - would be a 2 month job probably. I have a paper about an alternate algorithm, the authors put c code for this on github, but I cannot just lay my hands on it right now. I thought of checking that out when I have time.

tlunttil:
It would surely be quite a bit of work. Besides, I think internally (precision, performance, etc.) qhull is quite good, it's just that its C interface is a 'bit' awkward (C++ interface looks a lot better, but there are also other options for convex hull/Voronoi/Delaunay computations in C++).

allegroLeiden:
Yes, well, probably Lime should have been written in c++ to begin with, but I am quite glad personally that it wasn't, because I hate c++. ;)

qhull works when it works but I hate things which silently fail.

ParfenovS:
The problem with qhull is also that it is slow and it can take most of the time when constructing a grid. It would be great if there will be parallelized (at least partially) realization of triangulation.
Considering other questions, including C standard, I agree with @tlunttil.

allegroLeiden:
I think with the HCKB thing I will wait til Sebastien gives me the go-ahead to merge, because some of these conversions have already been done in master.

tlunttil:
One very weird thing about qhull is how slowly it works in LIME. The equivalent command line programs are a lot faster. I even tried dumping points from LIME to disc and running command line qhull in case LIME grids were somehow especially hard to triangulate, but even then there was a huge difference (a factor of 100 or more I think). I'm not sure if LIME calls qhull in some completely wrong way or if the command line program didn't actually do what I thought it would.

I even considered writing code to use qhull as an external program from LIME (which actually is what the documentation recommends in most cases). So I'd dump the point coordinates to a file, run qhull with system(), read facet dump from the output file, and reconstruct from that the needed data structures. Unfortunately writing such a piece of code didn't (doesn't) fell very exciting. However, I may do that anyway, because all the benchmarking stuff in #147 really needs a lot of points.

smaret:
> The gcc version on these (4.4.7) complains about c99 stuff, desires an option `-std=c99 or -std=gnu99`. I guess we could do that, and I don't have a really strong opinion here, but I would lean a bit in the direction against adopting c99 practices, unless there is a clear need.

Adding options in the Makefile may break compilation with some compilers. So I would rather keep plain ANSI-C (89?) until we implement a proper build system (with either cmake or autoconf/automake) than can detect which options each compiler supports.

smaret:
> I will wait till Sebastien gives me the go-ahead to merge, because some of these conversions have already been done in master.

From @tlunttil comments it seems that this PR is ready to be merged, so please go ahead.

smaret:
@tlunttil Could you please open a new issue about the qhull performances issues?

tlunttil:
Strictly speaking, those _Bool types should be removed if we are to stick to C89, because they only came in C99.

I'll open a new issue for qhull stuff.

imckstewart:
> Strictly speaking, those _Bool types should be removed if we are to stick to C89, because they only came in C99.

Ok - I don't get warnings with my older gcc but I guess that is because that version of gcc *recognized* _Bool, it just wasn't in the *standard*. I had not realized that. I can revert them all to ints if you like.

tlunttil:
gcc often supports (by default) some non-standard extensions. I guess that older gcc has a default language mode that is C89/C90 with some C99 (and other) extensions.

Personally I prefer the booleans and C99, but I don't have very a strong opinion about this. I remember I already saw booleans in some other (already merged) PRs, but I didn't think anything of it then.

allegroLeiden:
Oh, great. 17 modules requiring a manual merge. "I'm just going out for a while; I may be some time."

allegroLeiden:
Well, the merged stuff runs, but the results are not right, so I guess I won't be uploading today. I am about to knock off anyway, it is stinking hot right under the roof here. (Why are astronomy departments always right under the roof? Parkes, Leicester, Jodrell, Cape Town, now here. Only in Bonn was I on the ground floor.)

allegroLeiden:
I can't find any more places where I could replace stuff by HCKB.

tlunttil:
It looks like the changes had already been done in master.

allegroLeiden:
At lines 218 and 428 of raytrace.c there is a multiplication by md[0].norminv. '0' is surely wrong here but to use the formally correct molecule would require some fiddly changes and needs some thought, which is why I have left it. The effect is probably small, but this serves to flag that we have not yet, formally speaking, got rid of all the traps for multi-species use (issue #21).

tlunttil:
I'm not sure what exactly is the meaning of these norm/norminv things. Any comments?

allegroLeiden:
I thought it was a linewidth correction but it appears to be correction for BB spectrum. (shrug)

tlunttil:
I'll have closer look, I'm currently very confused as to what they are trying to accomplish.

tlunttil:
I don't think norm/norminv are doing anything useful at all. norm is just a scaling that is applied in one place and then removed in another with norminv. Maybe they were supposed to prevent over/underflows, but they're not really needed nor correctly used for that anyway.

Issue #151, opened 2016-08-02
Need to separate dust/knu from mol


allegroLeiden:
This is kind of a structural issue. At present, both dust and knu are pointer attributes of the mol field of grid point values. They are used both in making continuum and line images, and are also accessed when solving for the populations. In 'line' use they supply the relevant values at the frequency of each line being solved, but for continuum images they only have 1 value each, that of some representative frequency for the image.

The problem arises if one wants to supply precalculated population values. Both `continuumSetup()` and `levelPops()` free then re-malloc the `mol` pointer so as to get the correct size for dust/knu, which of course trashes any preloaded values of `mol.pops`. If dust/knu were in a separate structure, the above functions could resize this without affecting the pops values.

I'll put in a PR with a fix for this. It is preventing me implementing #92 fully.

allegroLeiden:
Fixed by #152

PR #152, opened 2016-08-08
Untangling - starting with the dust/knu thing

allegroLeiden:
There are a lot of changes in this single commit, but TBH I am about out of patience with nibbling away at major LIME issues over a space of months, having to waste time on copious amounts of repetition of the same mods and merges. I don't understand why some of you are worrying about the exactness of the results produced by this software when in my view it is at least 2 stages away from that point. We know there are problems with the algorithm; but the algorithm fixes are waiting on fixes for the code structure - the kind of thing which should have been sorted out before 1st release. First let's get the structure right, so it is not an enormous pain constantly to have to work around all the knots and peculiarities; then fix the algorithm; then by all means let's look at the numbers. The present PR is designed to make a leap towards fixing stage no. 1, i.e. code structure. There is more that could/should be done but hopefully the present PR accomplishes a major chunk of what was needed. The details of the changes are in the commit.

This addresses issue #151 BTW.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,4 +14,4 @@ example/lime_*.x
 example/vlog.txt
 
 # Ignore html doc files
-doc/_html
+doc/*.html
--- a/src/LTEsolution.c
+++ b/src/LTEsolution.c
@@ -10,28 +10,27 @@
 #include "lime.h"
 
 void
-LTE(configInfo *par, struct grid *g, molData *m){
+LTE(configInfo *par, struct grid *gp, molData *md){
   int id,ispec;
 
   for(id=0;id<par->pIntensity;id++){
     for(ispec=0;ispec<par->nSpecies;ispec++){
-      g[id].mol[ispec].nmol = g[id].abun[ispec]*g[id].dens[0];
-      lteOnePoint(par, m, ispec, g[id].t[0], g[id].mol[ispec].pops);
+      lteOnePoint(par, md, ispec, gp[id].t[0], gp[id].mol[ispec].pops);
     }
   }
-  if(par->outputfile) popsout(par,g,m);
+  if(par->outputfile) popsout(par,gp,md);
 }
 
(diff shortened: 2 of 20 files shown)


tlunttil:
This seems to have some overlap/conflicts with my PR that I submitted 15 minutes later...

allegroLeiden:
Um, yeah - do you have a cunning plan?

tlunttil:
It's probably easier to merge your PR and then redo mine on top of that. My PR doesn't have any structural changes apart from those related to the new lineshape/velocity integration stuff (and maybe we'll keep also the old scheme).

allegroLeiden:
I am going to force push in a minute to try to fix the wrapping of the commit message. Note to self: stop using %$#&^ pico as editor!

tlunttil:
Should I in turn have a look at this? We'll also need to think about how we'll eventually merge this and #153 (for example, should we also keep the old velocity integration option in the code).

allegroLeiden:
> Should I in turn have a look at this?

Yes please.

tlunttil:
I noticed that you read/write dummy variables to keep binary population files compatible with previous versions now that norm/norminv are no longer used. Is compatibility that important? I'm just thinking that if we don't drop the unnecessary fields from binary population files now, we'll probably never will...

allegroLeiden:
I had in mind to break backward-compatibility at release 2.0 (some time in the misty future). Hopefully at that point we would be able to throw away all the horrible CSV I/O altogether.

tlunttil:
This looks fine to me.

tlunttil:
Should we (well, @allegroLeiden) go ahead with the merge? This is big enough change to make a difference in writing many of the following PRs.

allegroLeiden:
I don't have many hours today before I head off to Lisbon but I'll try to bring this up to date with master. If I succeed then I'll merge it without more ado.

PR #161, opened 2016-08-31
New rotation

allegroLeiden:
This provides the necessary code to implement #63, which seems to have died of loneliness.

Note that I found my final prescription (comment of 9 Aug on the #63 thread) to be in error: the expression for `incMat` should have read

```
         ( c  0  s)
incMat = ( 0  1  0);
         (-s  0  c)
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -49,6 +49,7 @@ endif
 ##
 
 TARGET  = lime.x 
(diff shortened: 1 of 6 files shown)


smaret:
Looks good, thanks!

PR #168, opened 2016-09-12
User-friendly

lolMEvans:
I have tried to improve the user-friendliness of LIME by removing the need for multiple raytracings when
only using different image units and adding a new way of specifying what units to output each image. I have also added in a user-settable reference density to allow control over the grid weighting function outside of gridDensity(), which can be useful if the model density at r=par->minScale is significantly different than the features that are of interest.

I am new to git and GitHub so I apologise if I've made any mistakes with this pull request.


allegroLeiden:
Some of these changes certainly look useful but it may be a little while before we get to look at them - there is a bunch of stuff we want to merge into 1.7 asap, after that there will be a little breathing room.

Just as a general remark, it is better not to combine several changes in the one PR but to issue them in separate PRs.

allegroLeiden:
@lolMEvans do you feel like resubmitting the multiple-units stuff as a separate PR? I think it is a good idea and I would propose to merge it in after review. (E.g. it might be better to keep only 'units' in the user-settable image struct, moving the ones which the user doesn't set directly to the new struct 'imageInfo'. There is also now code in writefits.c which forces unit 5 to 2d fits.)

All the stuff to do with grid-point selection I would like to think about a bit more.

lolMEvans:
@allegroLeiden Yes I can do that. I'll just make sure it's up-to-date with the master branch and then I'll submit it. I've also been working on a more user-friendly feature of LIME where the filenames and extensions of all file are set automatically unless the user specifies a custom prefix- would you be interested if I submitted a pull request for this too?

PR #169, opened 2016-09-12
Add collScale parameter for scaling collisonal rates

migueldvb:
This is useful when considering collisional rates with a molecule for
which there are no values available in the literature. The collisional
rates from the LAMDA database can be scaled by the factor of the ratio
of molecular weights with the collision partner.

### DIFF
--- a/doc/usermanual.rst
+++ b/doc/usermanual.rst
@@ -477,6 +477,18 @@ If none of the three density-linked parameters is provided, LIME will attempt to
 
 .. code:: c
 
+    (double) par->collScale (optional)
+
+This parameter specifies an scaling factor used by LIME for the collisional rates
+read from the molecular data files defined in par->moldatfile. This is useful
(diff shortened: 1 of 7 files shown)


allegroLeiden:
Could one not achieve the same effect by modifying the appropriate molData file?

Speaking in general, I am not very enthusiastic myself about further complicating the parameter interface, in fact I look forward to the time when we can simplify it a bit. Of course this is not a rule set in stone, only my personal leanings. Other team members may have different ideas.

migueldvb:
I agree about keeping the interface simple. Maybe this would be useful for a small number of people.

allegroLeiden:
Michiel Hogerheijde: Better to leave collision rate scale to the user. Messing with the collision rates is NOT something LIME should do. It is at the user's own risk. We should not be seen to encourage this (although I do this myself often...).

PR #170, opened 2016-09-13
Remove duplicated variables defined in header file

migueldvb:


allegroLeiden:
Oops, I don't know how that happened!

Issue #173, opened 2016-09-15
User-defined temperatures of collision partners


migueldvb:
It is not supported in the code to have a different temperature distribution for each collisional partner. This could be useful to be able to define a temperature profile for the electrons when they are included as collisional partner and they are thermally decoupled from the neutral gas.

tlunttil:
It shouldn't be too difficult to allow the possibility of several temperature distributions. However, before writing a PR it's maybe better to wait until we've decided whether we're going to make changes to some of the related data structures. I think @allegroLeiden  may have some plans about that.

allegroLeiden:
Yes, I think it would be better to wait at least until after 1.7 before thinking about enhancements. The code structure ought to settle down by then.

migueldvb:
I think what is needed for the case of collisional partners with different temperature distribution is to be able to define collision rates as a function of position in the grid instead of reading them directly from the LAMDA datafiles. It should be possible to include a function in the model file that overrides the collision rates defined in the molecular datafile and at the same time maintaining backward compatibility.

tlunttil:
A problem with using a user-defined collision rate function is that it can cause a severe slowdown, because the function would be called many times when the stateq. system is constructed during iteration. With careless use of transcendental functions and `pow()`s that could easily become the main time sink.

One solution would be to precalculate the rate coefficients during grid construction and store them (in fact LIME used to do something like that), but that approach can easily lead to huge memory needs in case large models and molecules with many collisional transitions.

migueldvb:
Those are very good points. For some problems it may be necessary to have collision rates that depend on the position or some parameter other than the gas temperature, so it owuld be useful to have this functionality with a warning to users about possible problems.

tlunttil:
That should be easy to do, because the only place that needs to call the coll. coeffs function seems to be line 148 of stateq.c. If the change is done, it would probably also be a good idea to move the computation of the collision part of the stateq. matrix outside the ALI loop. That would reduce the number of calls to the coll. coeffs function by at least a factor of 5.

allegroLeiden:
Michiel Hogerheijde: Using different temperatures per collision partner is a must. However, I don't understand the argument about collision rates not being calculated for each grid point: they already have to be, since only the downward rates are given and the upward rates are calculated (they depend on T anyway; I may have gotten up- and down- confused here).

tlunttil:
[...]

PR #177, opened 2016-10-12
Add IR pumping read from an input file

migueldvb:
Added support for providing effective IR pumping rate coefficients in an
external file `girdatfile` for each radiating species. These coefficients can
be calculated from the contribution of cascading rotational levels
within vibration bands excited by IR solar radiation as described in
the case of comets by Bensch & Bergin 2004.  This effect is relevant
for cometary coma exposed to solar radiation.

### DIFF
--- a/doc/usermanual.rst
+++ b/doc/usermanual.rst
@@ -594,6 +594,17 @@ These last two parameters mostly replace the functionality of the older `par->ou
 
 .. _images:
 
+.. code:: c
+
+    (string) par->girdatfile[i] (optional)
+
+Path to the i’th data file containing the effective IR pumping rate
+coefficients that can be determined by the contribution of cascading
+rotational levels within vibration bands as in Bensch & Bergin 2004.
+This effect is relevant for cometary coma exposed to solar radiation.
+girdatfile is an array, so a different data file can be used for each radiating
+species.  If this parameter is not supplied the effect will be ignored.
+
 Images
 ~~~~~~
 
@@ -1314,6 +1325,7 @@ Appendix: Bibliography
 ----------------------
 
 -  Ade et al., A&A 576, A105 (2015)
+-  Bensch & Bergin, ApJ, 615, 531, 2004
 -  Brinch & Hogerheijde, A&A, 523, A25, 2010; see also
    http://www.nbi.dk/~brinch/lime.php
 -  Hogerheijde & van der Tak, A&A, 362,697, 2000
--- a/src/aux.c
+++ b/src/aux.c
@@ -192,8 +192,11 @@ The parameters visible to the user have now been strictly confined to members of
 
   } else {
     par->moldatfile=malloc(sizeof(char *)*par->nSpecies);
-    for(id=0;id<par->nSpecies;id++)
+    par->girdatfile=malloc(sizeof(char *)*par->nSpecies);
+    for(id=0;id<par->nSpecies;id++){
       copyInparStr(inpar.moldatfile[id], &(par->moldatfile[id]));
+      copyInparStr(inpar.girdatfile[id], &(par->girdatfile[id]));
+    }
 
     /* Check if files exist. */
     for(id=0;id<par->nSpecies;id++){
@@ -631,6 +634,7 @@ Eventually I hope readOrBuildGrid() will be unilaterally called within LIME; if 
       (*md)[i].lal = NULL;
       (*md)[i].lau = NULL;
       (*md)[i].aeinst = NULL;
+      (*md)[i].gir = NULL;
       (*md)[i].freq = NULL;
       (*md)[i].beinstu = NULL;
       (*md)[i].beinstl = NULL;
--- a/src/aux.c
+++ b/src/aux.c
@@ -107,7 +107,7 @@ parseInput(const inputPars inpar, image *inimg, const int nImages, configInfo *p
 The parameters visible to the user have now been strictly confined to members of the structs 'inputPars' and 'image', both of which are defined in inpars.h. There are however further internally-set values which is is convenient to bundle together with the user-set ones. At present we have a fairly clunky arrangement in which the user-set values are copied member-by-member from the user-dedicated structs to the generic internal structs 'configInfo' and 'imageInfo'. This is done in the present function, along with some checks and other initialization.
   */
 
-  int i,j,id,status=0;
+  int i,j,id,status=0,numGirDatFiles;
   double BB[3],normBSquared,dens[MAX_N_COLL_PART],r[DIM];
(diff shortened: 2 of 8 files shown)


imckstewart:
It's great that you are proposing these new features for Lime but you may need to be a bit patient! My priorities at the moment at least are (i) to merge in the structural and other fundamental PRs for 1.7, (ii) to work on the python/casa interface. I don't think any of us central developers will have much time to look at new functionality before 1.7 at least. I had hoped to get that out by end of Nov, am currently waiting on @tlunttil who is busy with non-Lime matters.

imckstewart:
I am also @allegroLeiden by the way. :smile:

tlunttil:
This PR looks safe to me, though, and it doesn't seem to touch any parts that are impacted by the structural/bugfix PRs that are in the pipeline. So I wouldn't think that including this in 1.7 would be too hard.

Today I should finally have time to look at merging #153.

imckstewart:
Well, (i) I don't understand the physical justification, (ii) haven't got time to read up on it now, and (iii) the PR involves yet another file/parameter added to Lime's already bloated input. The whole thing is not clear to me and I would prefer to get my head around it before giving it my vote. And just at the moment I don't have the time to do that.

tlunttil:
It seems that there's a clear need for allowing 'extra' (de)excitation input into stateq calculations due to collisions with electrons, unmodelled (in LIME) radiative pumping and maybe some other things that nobody has suggested yet. However, instead of adding all separate possibilities one by one, we should think of a general enough framework to allow (almost) all possibilities. Maybe something like @migueldvb suggested in #173, i.e. possibility of a user-defined function giving (additional) transition rates as a function of location.

migueldvb:
Thanks for your comments. These coefficients are used for molecules in cometary come that are in                          
fluorescence equilibrium, as they are excited by solar radiation pumping and                    
subsequently undergo spontaneous decay to the rotational levels within the                      
ground vibrational state.  Fluorescence equilibrium is important only in the                    
outer, collisionless coma for certain molecules. I think that this kind of                      
mechanism is useful to have as there are several groups that use LIME for                       
modeling comets but it could be part of a more general implementation.

allegroLeiden:
Sounds like a good idea @tlunttil . Timing is a bit the problem for me at the moment: I've a deadline to produce a CASA-callable Lime by the end of Nov, so I am working flat out on C/python code at the moment. Hoping to slip 1.7 into some cracks somewhere but anything else will have to wait a bit.

allegroLeiden:
Michiel Hogerheijde: IR pumping again is something relevant for comets. THIS IS NOT EASY TO IMPLEMENT! Some molecules go through pre-dissociative states due to IR pumping, so the spectroscopy here is tough! It's a PhD project in itself to fully adapt LIME for comets. This should not be attempted through a pull request, but rather full development of a comet-ready LIME version. Important not to do a half job here.

migueldvb:
I agree with Michiel's comment. Here I only wanted to allow for giving additional excitation rates in the statistical equilibrium equation that could be useful for other mechanisms as well. The effective pumping rates do not have to be calculated in LIME necessarily. Perhaps this can be done in an external package and the pumping rates read in the file containing the model. This calculation needs to query spectal catalogs in IR wavelengths like HITRAN or GEISA and it is easier to implement in a language like python where the query tools are available.

migueldvb:
Resolved merge conflicts in this branch.

allegroLeiden:
- The way you have written it at present, if the user supplies any moldat files, the code assumes there will be at least an equal number of girdat files. This is not backward compatible. You need perhaps a bit more code starting somewhere around line 188 of aux.c (in your branch). The logic of this should be as follows:

```
  // Determine numGirDatFiles in a way similar to lines 184-186.
  if(numGirDatFiles<=0)
    par->girdatfile = NULL;
  else if(numGirDatFiles!=par->nSpecies){
    // fail with error
  }else{
    par->girdatfile=malloc(sizeof(char *)*par->nSpecies);
    for(id=0;id<par->nSpecies;id++){
      copyInparStr(inpar.girdatfile[id], &(par->girdatfile[id]));
    }
  }
```
Of course you would extract the girdatfile additions from present lines 195-198.

- The stuff you added to `readMolData()` should go in a separate function which can then be called from `molInit()` if(par->girdatfile!=NULL).

- There should be a similar test before adding the gir stuff to the matrix in `getmatrix()`.

allegroLeiden:
@migueldvb I'd like to try what @tlunttil suggested on thread #173 on 3rd Oct last year - i.e. to move the unchanging-value matrix calculations outside the loop at line 45 in stateq.c. You might prefer to wait until that is done, and put your own additions in the same place.

migueldvb:
Those are good points. I'm in a middle of an observing run at the moment, I'll try to fix implement the changes next week.

allegroLeiden:
@migueldvb , it doesn't pass the travis build.

migueldvb:
Oops, I'll take a look.

migueldvb:
Travis checks are passing now. I have included `configInfo *par` as an argument to `getMatrix` in order to test for girdatfile being defined.

allegroLeiden:
I guess this is ok... I think I would have put lines 90-96 of stateq.c into getFixedMatrix() myself, since they only need to be calculated once. You'd have to add another gsl_matrix* argument to getFixedMatrix() to return the result, and modify line 161 to add it in to reduc, and add a bit more mess... it is arguably neater the way you have it, and not many people I guess will use this, but those who do will get slower running than they need to, because this bit of code is fairly time-critical. I can pass on it though and leave other concerns for future generations. There are other things I want to do with LIME which are banking up a bit.

You also could have tested for md[ispec].gir==NULL rather than passing in configInfo*, but it works as it is, so leave it.

A very minor point: do you want to indent lines 105 and 114 of stateq.c by 2 spaces each?

If you can do that, then merge in the latest changes to master via
```
  git pull upstream master
```
etc, I will merge the PR.

allegroLeiden:
Miguel I'd appreciate it if you could make some haste with this. There are other things I need to do with LIME which I am holding up until I can merge your PR. The alternative is if I forge ahead, then you have more to merge, which introduces more problems, until the whole thing starts looking like an infinite loop.

migueldvb:
Thank you, that sounds great. I have corrected the indentation in stateq.c and merged this branch on top of the latest changes in the master branch. The CI checks are running at the moment.

allegroLeiden:
Hi Miguel, many thanks for that. Sorry to hurry you. The git log looks like it might be screwy, so I might have to merge it 'off-piste', but I will take it from here anyway.

migueldvb:
I thought that this could merge cleanly. Sorry about the git problem.

Issue #162, opened 2016-09-03
Funky looking cells in the fits image


cleevesli:
Hey all,

I cloned yesterday's version of LIME and have been putting in my own model reading code bits back in, and ran identical models before and after the update to compare. I've noticed in the most recent versions (both the v1.6 commit yesterday and a few days ago) that there are now giant cells in the high velocity line channels for both the red and blue shifted emission (see attached images). Any thoughts on what could be causing this? The vtk grid files as viewed with ParaView look fine, and the fluxes for the circumstellar disk emission (the center bowtie part) are very similar to the previous version. 

Best,
Ilse

<img width="477" alt="screen shot 2016-09-03 at 6 51 42 am" src="https://cloud.githubusercontent.com/assets/6672240/18224658/f5ab1570-71a2-11e6-8e44-0a55aa7e9858.png">

<img width="458" alt="screen shot 2016-09-03 at 6 52 06 am" src="https://cloud.githubusercontent.com/assets/6672240/18224659/f9892056-71a2-11e6-93f7-f19e905153de.png">

cleevesli:
I also just noticed if I do an extremely narrow stretch on the color bar range that the background is a little strange looking too, with the black parts being 1e-30 and the grey being exactly 3.21e-5:

<img width="710" alt="screen shot 2016-09-03 at 6 57 18 am" src="https://cloud.githubusercontent.com/assets/6672240/18224682/b3495fec-71a3-11e6-971a-b936874f4291.png">

Best,
ilse

smaret:
Hey Ilse !
Do you expect any gas emission in these high velocity channels or dust emission only?  The way the dust density computation has changed in v1.6 (see the discussion in #128), so you may what to check this.

cleevesli:
Hello Sebastian!

Last night I started backtracking through commits to see where the issue first appears and was able to trace it back to the large August 1st commit: 

https://github.com/lime-rt/lime/commit/e9956e8d409e21a26144b5e306bdd9a519c5af2b

It is not present in the July 20th commit:

https://github.com/lime-rt/lime/commit/9eaa28365db1403277606d20aaf7c97bdd053092

And I haven't yet gotten the July 21st commit running with my chemical model inputs, but August 1st is definitely not working. 

Regarding line vs. dust, it's definitely in the line emission, and to check I re-ran the model without dust and the artifacts in the fits image are still there.

I don't know exactly where the issue appears within the August 1 version, but I'll start backtracking through that one next..

Best,
[...]

PR #183, opened 2016-11-08
Hard-wired DIM==3 removed from ray-through-cells algorithm.

allegroLeiden:
I have made the raythrucells algorithm generic (i.e. no longer hard-wired to DIM==3), and the main reason for doing this (apart from promoting those noble software goals, elegance and robustness) was to allow the same code to be used in the dims==2 situation of following image raster lines through the 2D projected cells in raytrace(), which I had hoped would address #162. However I have just this instant realized that I have shot myself in foot here because I have arranged matters such that the number of dimensions is set via a macro in the new header file dims.h. Since I have made use of structs with array members with sizes set by DIM, which I did because it saves an awful lot of mallocing, I cannot of course also use them in the same executable for dims==2. So cut-and-paste in here whatever swearing you deem appropriate. I will have to think about this some more. In itself it is a reasonable (if slightly sterile) PR but it doesn't advance towards solving #162.

### DIFF
--- a/src/dims.h
+++ b/src/dims.h
@@ -1,0 +1,9 @@
+#ifndef DIMS_H
+#define DIMS_H
+
+#define DIM 3
+#define N_DIMS DIM
+
+#endif /* DIMS_H */
+
+
--- a/src/frees.c
+++ b/src/frees.c
@@ -9,7 +9,8 @@
 
 #include "lime.h"
 
-void freeConfig(configInfo par){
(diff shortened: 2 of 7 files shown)


allegroLeiden:
I suppose I could pass in numDims to all the raythrucells functions which make use of the number... then if I dimension all the arrays to 3 but pass in 2... that ought to work.

PR #185, opened 2016-11-09
Remove all but user-settable values from struct image.

allegroLeiden:
PR #77 did this for the inputPars. I am not sure why I did not also do the same job at that time on the user-settable image parameters, but I am doing so now.

The motivation for doing this is the same as for #77. An additional reason is that it allows easier integration of Lime into [Artist.](http://youngstars.nbi.dk/artist/Welcome.html)

### DIFF
--- a/src/aux.c
+++ b/src/aux.c
@@ -15,8 +15,39 @@ TODO:
 #include <float.h>
 
 
+/*....................................................................*/
 void
-parseInput(inputPars inpar, configInfo *par, image **img, molData **m){
+mallocInputPars(inputPars *par){
+  int i;
+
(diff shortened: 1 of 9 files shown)


allegroLeiden:
My aim to add a facility to Lime to allow it to read parameters and functions from a python model file provides an additional reason for having completely independent memory areas for the user-settable vs. the internal parameters. For such code I envisage (i) putting run() into a separate module and (ii) writing a python-specific main() in a new 'pymain.c' or whatever module which would pass 'inputPars' and `image` objects to run(). It is not so easy to supply `char*` via python so one could not any more just copy these pointers from one struct to another as in present Lime. (Or so I have been assuming anyway.)

Issue #181, opened 2016-10-31
Poor control of convergence


tlunttil:
Currently the number of global iterations in LIME is hardwired in lime.h to 16 and there is no checking whether the solution has converged by then. 

A problem linked to this is that the ray directions are not the same for each iteration. (There is also an extra source of randomness in multithreaded computations due to updating the level populations on the fly). This mixes the effects of Monte Carlo noise and iterations, which makes it very difficult to monitor convergence.

allegroLeiden:
PR #92 includes a new parameter which allows the user to set the number of iterations. However of course I agree with the rest of the issues.

allegroLeiden:
This was fixed in 1.7.

Issue #160, opened 2016-08-30
Remove scaling of rays per point according to the number of neighbours?


tlunttil:
Currently the number of rays sent from each grid point is proportional to the number of its Delaynay neighbours. There doesn't seem to be any good reason for that and furthermore, the current implementation has some problems. Some buffers are scaled according to a (conservative) fixed estimate for the number of neighbours a grid point can have (essentially max_phot/ininphot). In most cases this just wastes memory, because the buffers are unnecessarily large. However, there is no check for the true number of neighbours, and it is, at least in theory, possible that the buffers are too small and the program crashes.

allegroLeiden:
Agreed.

tlunttil:
This was done in #191.

PR #187, opened 2016-11-10
Partial fix to #181 and a bug fix in photon()

tlunttil:
This PR saves the random seeds and resets RNGs so that each iteration will have use the same set of rays. This should allow better monitoring of convergence in a yet-to-be-written PR. This PR also includes a bug fix/improvement to photon(), where the in/out split was not used quite correctly.

### DIFF
--- a/example/model.c
+++ b/example/model.c
@@ -27,6 +27,7 @@ input(inputPars *par, image *img){
   par->antialias                = 4;
   par->sampling                 = 2; // log distr. for radius, directions distr. uniformly on a sphere.
   par->nSolveIters              = 14;
+  par->resetRNG	                = 0;
 
   par->outputfile               = "populations.pop";
   par->binoutputfile            = "restart.pop";
--- a/src/aux.c
+++ b/src/aux.c
@@ -798,7 +798,7 @@ While this is off however, other gsl_* etc calls will not exit if they encounter
         }
         halfFirstDs = malloc(sizeof(*halfFirstDs)*max_phot);
 
-#pragma omp for
+#pragma omp for schedule(dynamic)
         for(id=0;id<par->pIntensity;id++){
 #pragma omp atomic
           ++nVerticesDone;
--- a/src/aux.c
+++ b/src/aux.c
@@ -703,6 +703,7 @@ levelPops(molData *md, configInfo *par, struct grid *gp, int *popsdone, double *
   struct blendInfo blends;
   _Bool luWarningGiven=0;
   gsl_error_handler_t *defaultErrorHandler=NULL;
+  int RNG_seeds[par->nThreads];
 
   nlinetot = 0;
   for(ispec=0;ispec<par->nSpecies;ispec++)
@@ -730,7 +731,7 @@ levelPops(molData *md, configInfo *par, struct grid *gp, int *popsdone, double *
 
     for (i=0;i<par->nThreads;i++){
       threadRans[i] = gsl_rng_alloc(ranNumGenType);
-      gsl_rng_set(threadRans[i],(int)(gsl_rng_uniform(ran)*1e6));
+      RNG_seeds[i] = (int)(gsl_rng_uniform(ran)*1e6);
     }
 
     calcGridCollRates(par,md,gp);
@@ -777,12 +778,14 @@ While this is off however, other gsl_* etc calls will not exit if they encounter
 
       calcGridMolSpecNumDens(par,md,gp);
 
+
       nVerticesDone=0;
       omp_set_dynamic(0);
 #pragma omp parallel private(id,ispec,threadI,nextMolWithBlend) num_threads(par->nThreads)
       {
         threadI = omp_get_thread_num();
 
+        gsl_rng_set(threadRans[threadI],RNG_seeds[threadI]);
         /* Declare and allocate thread-private variables */
         gridPointData *mp;	// Could have declared them earlier
         double *halfFirstDs;	// and included them in private() I guess.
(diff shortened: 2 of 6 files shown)


tlunttil:
Note that in this PR resetting the RNG for each iteration is always used. Maybe that should be a user option?

tlunttil:
A further comment: the potential race condition problem caused by updating the level populations on-the-fly had already been solved as a side product of @allegroLeiden 's PR #152.

allegroLeiden:
> Note that in this PR resetting the RNG for each iteration is always used. Maybe that should be a user option?

That would probably be better - at least for now. For Lime-2.0 we could look forward to boiling away all the excess alternatives (rubs hands together).

A couple of other comments:
- `NITERATIONS` is of course redundant now but I was thinking of transmogrifying it to `MAX_N_SOLVE_ITER` and defaulting `par.nSolveIters` to this number. Thoughts?
- Why the added space in the `DS_mask_temperatures` macro?

tlunttil:
Yes, I started thinking that making it an optional thing may be better for the time being. So should I put another parameter in configInfo?

The other comments:

- Ah, ok. I thought it had NITERATIONS had just been forgotten. If it was left there by design, I could put it back.
- I think I must've dropped something on my keyboard...

allegroLeiden:
cat_it

allegroLeiden:
I guess another parameter is inevitable... I look forward to a grand cull one day.

allegroLeiden:
I just mentioned this PR to Michiel, and he made an interesting suggestion: that the number of photons per grid cell needs at some level to be adaptively altered; and that the criterion for increasing the number of photons for any grid cell (it would unfortunately require that info to be kept for each grid cell) would be some function of the noisiness of the population time series for that cell. He suggested a 'fixed-seed' phase to arrive quickly at a solution, which will probably be quite a good one for the majority of grid cells, followed by a 'random-seed' phase in which the adaptive stuff would take place. I don't propose that you should go away and code such up, but I'd be interested in your views on it, because it sounded a good idea to me.

tlunttil:
Yes, I think that is a good idea. So good that I think I once proposed something similar myself ;) (And all this is of course also related to #160.)

Making it work (well) requires a complete reworking of convergence monitoring, which I think is going to have to wait a while.

tlunttil:
I made the RNG reset thing user-settable and defaulting to off and returned `NITERATIONS `to lime.h

I also seem to have removed unused parameter `OtoP` from lime.h which may have meant ortho-to-para ratio. I guess no-one will miss that?

allegroLeiden:
> I guess no-one will miss that?

Let them eat cake.

allegroLeiden:
> So good that I think I once proposed something similar myself ;)

Quite possibly, somewhere in the wilderness of github-land. I thought however that your proposal was to adaptively weight the photon *directions*, which is a bit different.

tlunttil:
> Quite possibly, somewhere in the wilderness of github-land. I thought however that your proposal was to adaptively weight the photon directions, which is a bit different.

Hmph, you've thwarted my attempt to usurp priority. 

My idea (I think we've only discussed this in emails) is basically to improve sampling of J by making the angular sampling better in some of the cells. The simplest option is of course to just use more rays with isotropically distributed directions; that adaptive direction stuff is a further step.

However, my idea to look for noisiness in contributions from rays from different directions (during one iteration) to select where to refine the sampling, instead of using changes in level populations between iterations. It seems to me that looking at different iterations mixes the effect of MC noise and convergence and it would be hard to know which of these is causing the variations. Although if I understood correctly, Michiel's idea would be to start the scheme only after reaching a semi-converged solution, so variation would indeed be mostly caused by MC noise due to random ray directions (and frequencies, as long as we're still doing that with MC),

allegroLeiden:
> variation would indeed be mostly caused by MC noise due to random ray directions

Well ok, but then what is 'MC noise' when you think about it? Imagine 2 grid points: point A receives fairly isotropic radiation, but inward radiation to point B is dominated by a bright source nearby in the model. Would one not expect that separate iterations with random photons would produce a fairly quiet time series for A but a jumpy one for B? Ideally we would in that case, for B, put more photon directions towards the bright source. An adaptive way to do that would be fine but a simple brute-force method in which the photon distribution was kept isotropic but the number was increased would also work, more robustly if not as efficiently.

However I don't think this is something we could feasibly jam into 1.7.

Sorry if I forgot the previous discussion - I'm an old bugger, I forget things.

tlunttil:
> Would one not expect that separate iterations with random photons would produce a fairly quiet time series for A but a jumpy one for B?

Yes, I agree, at least after a few iterations. In the early phase I could imagine there being a lot of jumpiness even in an isotropic field due to convergence (i.e., even with perfect sampling J would change significantly from one iteration to another, but with MC noise on top of that it could be hard to tell what is causing what).

I completely agree that this isn't something for 1.7. Indeed, I think a more important short-term goal would be to get some sort of convergence monitoring. Currently the iteration is just stopped after a set number of rounds without any checks (and even that may be hard to do in time for 1.7). This sampling stuff is a different, but related, issue.

> Sorry if I forgot the previous discussion - I'm an old bugger, I forget things.

I'm not sure if I ever discussed this much with you or possibly someone else, or if they were just some private thoughts. Anyway, my comment about priority wasn't entirely serious ;)

tlunttil:
And of course there's still that strange 'scale the number of rays according to Delaunay edges' thing going on (issue #160). I think I'll get to that next.

allegroLeiden:
I am working on changing choice of next edge from nearest-to-photon-direction to nearest-to-photon-track at the mo. PR will come shortly.

PR #188, opened 2016-11-11
Detection and promotion of geometric sink points.

### LINKED ISSUE #101: External grid points can fail to be labelled as sink points

allegroLeiden:
Fixes #101.


tlunttil:
Shouldn't the sink check for be also done in delaunay calls during smoothing iterations? Smoothing moves points around, so isn't it possible that a non-sink point ends up belonging to the convex hull?

allegroLeiden:
It doesn't matter during the smoothing (I don't think so anyway), thus it only needs to be done in the final delaunay *after* the smoothing.

tlunttil:
Yes, ok. Somehow I thought that smoothing happened later that the delaunay call with sink checking, but now I see that isn't true.

allegroLeiden:
It used to be that way.

Issue #189, opened 2016-11-11
Faulty 1st step in photon propagation


allegroLeiden:
At line 280 in the present photon.c the code executes `continue` to progress to the next grid point (the first grid point away from the photon start point). However this is done without setting `here` to this next grid point, so `here` is still set to the start point at the following (i.e. 2nd) pass through the propagation loop. One can fix this easily enough by inserting a line `here=there;` before the `continue` but this then exposes another potential problem in that theoretically the new `here` may be a sink point, at which the the propagation is supposed to terminate.

I will fix this in an imminent PR which has the main aim of changing the choice of next edge from that being closest (with some jitter) to the photon direction to that which keeps closest (ditto jitter) to the photon trajectory, while still going forward. What I propose to do is add the `here=there;` line as above but also move the test for `gp[here].sink` from the end of the loop to the beginning. As photon() is presently called this amounts to a surplus test for the first grid point, but it looks like a small cost to retain simplicity.

tlunttil:
I remember that the test for sink points has been oscillating between the beginning and the end of the loop for a couple of times now and it has also changed between testing `gp[here].sink` and `gp[there].sink`. But I think we're nearing convergence ;)

allegroLeiden:
Plus ça change...

PR #190, opened 2016-11-11
Photon keeps near nominal track rather than to direction.

allegroLeiden:
The function nextEdge() has been changed to prefer an edge which keeps the path through the model near the nominal track of the photon rather than an edge whose direction is as near as possible to the photon direction.

@tlunttil you are going to hate this because it runs 20% slower, throwing away all your nice speed gains.

I included a variant with some precalculation, which is faster, but still not as good as pre-PR.

Having written this thing I am starting to wonder how necessary it is! After all, the sampling of direction-frequency space is so sparse, perhaps it is not so important to reduce the fuzzy sampling of the model which the old algorithm implemented. This would probably only grow to importance if we implement adapative photon directions...

I also included a fix for #189.

### DIFF
--- a/src/aux.c
+++ b/src/aux.c
@@ -764,6 +764,7 @@ Pointers are indicated by a * before the attribute name and an arrow to the memo
   }
 }
 
+/*....................................................................*/
 void
 levelPops(molData *md, configInfo *par, struct grid *gp, int *popsdone, double *lamtab, double *kaptab, const int nEntries){
   int id,iter,ilev,ispec,c=0,n,i,threadI,nVerticesDone,nItersDone,nlinetot;//,conv=0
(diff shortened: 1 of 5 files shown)


tlunttil:
Hmm, it seems to me that you're maximising the cos between the line from the starting position to the new position and inidir, which isn't exactly the same as minimising the distance between the new position and the line defined by inidir.

allegroLeiden:
No, but it was easier, and I surmised almost as good. I think finding the perpendicular distance would require an additional vector scalar multiplication, subtraction and squaring.

tlunttil:
But you wouldn't need sqrt and division, which are a lot more expensive.

PR #191, opened 2016-11-13
Number of rays no longer scales with edges

tlunttil:
The number of rays per grid point is now set by `RAYS_PER_POINT` in lime.h, currently set to 200. The number 200 is arbitrary, but the previous `ininphot 9` resulted in roughly similar (perhaps slightly smaller) number of rays on average.

Each grid point has (already had) an individual `nphot` attribute, which is currently set to `RAYS_PER_POINT` for every non-sink grid point. Currently that is not needed, but it can be useful later when adaptive ray sampling is implemented.

Sampling of frequencies was to some extent tied to the previous 'scale the number of rays according to edges' scheme. I didn't bother trying to recreate that, because the whole frequency integration part will probably be redone soon. Instead, I'm using a very simple uniform sampling, which is more of a place holder rather than something that I'm proposing for real use.

### DIFF
--- a/src/aux.c
+++ b/src/aux.c
@@ -847,9 +847,7 @@ While this is off however, other gsl_* etc calls will not exit if they encounter
           stat[id].pop[ilev+md[0].nlev*4]=gp[id].mol[0].pops[ilev];
         }
       }
-
       calcGridMolSpecNumDens(par,md,gp);
-
 
       nVerticesDone=0;
       omp_set_dynamic(0);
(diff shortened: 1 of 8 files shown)


allegroLeiden:
I am a bit uneasy with the uniform frequency sampling - a pseudo-random sampling would have maybe been better - but as you say, it is just a placeholder. I'll merge this and we'll see what sort of output it gives.

tlunttil:
I'm currently trying some different frequency sampling methods. In particular, I'm implementing a scheme which traces different frequencies along a single ray simultaneously, so that some of the geometry calculations can be reused. Unfortunately the code structure isn't quite optimal for doing that efficiently. I'll make a new PR if something useful emerges.

PR #194, opened 2016-11-17
Calculate density as weighted sum in popsout

migueldvb:
This method is used to calculate the number density of radiating species
in calcGridMolDensities and it would make sense to do the same here
to return the relative abundance nmol/dens that was provided by the
user.

### DIFF
--- a/src/popsout.c
+++ b/src/popsout.c
@@ -26,7 +26,7 @@ popsout(configInfo *par, struct grid *g, molData *m){
   fprintf(fp,"# Column definition: x, y, z, H2 density, kinetic gas temperature, molecular abundance, convergence flag, pops_0...pops_n\n");
   for(j=0;j<par->pIntensity;j++){
     dens=0.;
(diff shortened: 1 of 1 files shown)


allegroLeiden:
Technically you are right, and I will merge this. I should mention though that I have a low opinion of popsout - reading and writing to ascii files in this way is pretty horrible - and I have not exerted myself to keep it up to date with recent changes. That's one of the reasons I developed the comprehensive grid IO of PR #92. A lot of people seem to hate FITS these days but I arranged things so that it ought to be easy to add other backends, to HDF5 for example. Anyway at some stage, maybe in a year or so, we will release Lime-2.0 which will not preserve backward compatibility and in that release I will be pushing for a lot of superseded dead wood such as popsin/out to be discarded.

migueldvb:
This may be useful for people that are reading the densities directly from the populations text file. It is nice to hear that this will be replaced by the new IO system.

PR #198, opened 2016-11-30
Check if density is non-zero at r = par->minScale

migueldvb:
It could happen that the density function is not defined at the origin. Comet models have generally a singularity at r=0 and minScale is defined above the surface. It is safer to evaluate the function at r=minScale that has been supplied by the user.

### DIFF
--- a/src/aux.c
+++ b/src/aux.c
@@ -56,7 +56,7 @@ The parameters visible to the user have now been strictly confined to members of
 
   int i,j,id;
   double BB[3],normBSquared,dens[MAX_N_COLL_PART],r[DIM];
-  double dummyVel[DIM];
+  double dummyVel[DIM],dummyScale;
   FILE *fp;
   char message[80];
   _Bool doThetaPhi;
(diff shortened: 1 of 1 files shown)


allegroLeiden:
Hmm well in my view, the user ought to know better than to write a model with a singularity. I prefer the code to refuse to function in such situations rather than to attempt to do the user's thinking for them. The lines you have changed would avoid an error at r=0, which will cover most of the cases, yet still leave the code open to a potential overflow, since there is no mechanism to force grid points to avoid |r|<minScale (nor should there be, since singularities can be anywhere). I have been thinking rather of moving this section of code in the opposite direction: to keep the r=0 and test specifically for an overflowed result.

migueldvb:
I thought that the default sampling algorithm selects grid points randomly in log(r) between minScale and radius. That is what the user expects as these can be the inner and outer boundaries of the model. In that case it would probably be safer to make this check somewhere in the range defined by the user.

allegroLeiden:
> I thought that the default sampling algorithm selects grid points randomly in log(r) between minScale and radius.

Only for par->samplingAlgorithm==0 and  par->sampling==0 or 2.

> In that case it would probably be safer to make this check somewhere in the range defined by the user.

Wouldn't it be no more trouble to the user (and much better practice) simply to avoid singularities in their model?

My bottom line here is that Lime is public software. Some of us have been working pretty hard over the past year or two to drag it by the scruff of its neck up to, or a bit closer to, the standard one expects of public software. It is not like the situation with one's own private software, where one can hard-wire as many caveats, short cuts and special cases in as one pleases. With public software one trades a certain amount of convenience for (at least eventually) a gain in robustness and support. In the present case, the call to velocity does what it is meant to do, i.e. tests to see that the function is present, regardless of whether the result overflows or not. With some modification of the density function to test for overflows, it will also serve its intended function. I am very much against hard-wiring in assumptions about the locations of singularities, and I don't see that allowing the user to tell Lime where there is a singularity is much less work for the user than to avoid singularities in the first place. If the user wants to supply a model with a singularity at 0, the code changes I am envisaging would warn about that, but not exit with an error; the risk of Lime choosing a grid location in a place which will cause an overflow during the subsequent solution phase would then be caveat emptor. I don't think there is any other way to arrange matters which is consistent with public code standards.

allegroLeiden:
Invalidated by #214.

PR #202, opened 2016-12-08
Add name of variables in header of populations file

migueldvb:
Although the ascii output is going to be deprecated it is still used to read the output populations. This changes the header to use whitespace delimiter for column names as in the rest of the file. The population file can then be read as CSV file by data analysis tools such as python pandas without having to specify
column names manually.


allegroLeiden:
Sounds sensible, but I am kind of lying on the floor with my tongue hanging out at the mo. 3 more work days then I go on holiday for 3 weeks. Maybe I'll have the energy to address these lime issues after that.

PR #207, opened 2017-01-09
Reinstates subtraction of CMB at the end of each ray trace.

### LINKED ISSUE #206: Quality assesment
I have people emailing me with questions about what is wrong with LIME. I will pass this on to the community since I have had little to do with the recent updates. I have no idea of what is going on, but something is clearly badly broken. 

Here is a continuum image of a very simple model, a box of constant density and temperature. This is what I get from LIME 1.41. The size is correct and the intensity is correct too:
figure_1

The exact same model in version 1.7. Notice the factor of ~100 in peak intensity.
figure_2

This is rather concerning. Has anyone actually bothered to test the code?

allegroLeiden:
In traceray() and traceray_smooth() the subtraction of local_cmb from the ray intensities has been reinstated (it was removed from 1.6 since Tuomas and I did not understand its purpose).

This should fix #206.

### DIFF
--- a/src/aux.c
+++ b/src/aux.c
@@ -173,11 +173,6 @@ The parameters visible to the user have now been strictly confined to members of
   while(i<MAX_N_HIGH && par->gridDensMaxValues[i]>=0) i++;
   par->numGridDensMaxima = i;
 
-  /* Check that the user has supplied the velocity function (needed in raytracing unless par->doPregrid). Note that the other previously mandatory functions (density, abundance, doppler and temperature) may not be necessary if the user reads in the appropriate values from a file. This is tested at the appropriate place in readOrBuildGrid().
-  */
-  if(!par->doPregrid || par->traceRayAlgorithm==1)
-    velocity(0.0,0.0,0.0, dummyVel);
-
   /* Calculate par->numDensities.
   */
   if(!(par->doPregrid || par->restart)){ /* These switches cause par->numDensities to be set in routines they call. */
@@ -493,6 +488,11 @@ LIME provides two different schemes of {R_1, R_2, R_3}: {PA, phi, theta} and {PA
     else
       par->nContImages++;
   }
+
+  /* Check that the user has supplied the velocity function (needed in raytracing unless par->doPregrid). Note that the other previously mandatory functions (density, abundance, doppler and temperature) may not be necessary if the user reads in the appropriate values from a file. This is tested at the appropriate place in readOrBuildGrid().
+  */
+  if(par->nLineImages>0 && (!par->doPregrid || par->traceRayAlgorithm==1))
+    velocity(0.0,0.0,0.0, dummyVel);
 
   /* Allocate moldata array.
   */
--- a/src/grid.c
+++ b/src/grid.c
@@ -5,6 +5,8 @@
  *  Copyright (C) 2006-2014 Christian Brinch
  *  Copyright (C) 2015-2016 The LIME development team
  *
+TODO:
+	- In readOrBuildGrid(), test for the presence of the 5 mandatory functions (actually 4, since velocity() is already tested in aux.c:parseInput() ) before doing smoothing.
  */
 
 #include "lime.h"
--- a/src/lime.h
+++ b/src/lime.h
@@ -69,11 +69,12 @@
 #define PI                      3.14159265358979323846	/* pi	*/
 #define SPI                     1.77245385091		/* sqrt(pi)	*/
 #define maxp                    0.15
-#define NITERATIONS 		16
+#define NITERATIONS             16
 #define MAX_RAYS_PER_POINT      10000
-#define RAYS_PER_POINT		200
+#define RAYS_PER_POINT          200
 #define minpop                  1.e-6
 #define eps                     1.0e-30
+#define IMG_MIN_ALLOWED         1.0e-30
 #define TOL                     1e-6
 #define MAXITER                 50
 #define goal                    50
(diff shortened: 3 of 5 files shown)


tlunttil:
Thanks for doing this, I was planning to fixing this myself but then things got sidetracked...

I noticed earlier that in 1.7 there was still something fishy going on near the model-radius circle, namely that there were some points on/near the circle that were apparently clamped (falsely?) to eps and caused some strange looking things in interpolation. Were 'other changes' meant to address this?

tlunttil:
Another thing that was mentioned earlier: I'd prefer tagging 'outside the model' pixels during raytracing (with NaNs for example). Then during fits output NaNs could be caught and replaced with IMG_MIN_ALLOWED or something else (possibly define IMG_VALUE_OUTSIDE or something). But maybe that is better left to another PR.

allegroLeiden:
> I noticed earlier that in 1.7 there was still something fishy going on near the model-radius circle, namely that there were some points on/near the circle that were apparently clamped (falsely?) to eps and caused some strange looking things in interpolation. Were 'other changes' meant to address this?

No, because I haven't seen such things myself. What model did you use to see these effects?

> I'd prefer tagging 'outside the model' pixels during raytracing (with NaNs for example).

This is probably an academic discussion because I suspect the astronomers are going to have their way on this issue. You know how that goes. However FWIW I guess the root issue is the somewhat philosophical question of whether we consider areas outside the model radius to be *terra nullius* or whether we consider them as comprising part of the model, just with zero density. I began as a partisan of the former view but I have moved to the latter. In this case the r>R areas ought really to start off anyway with the CMB value of brightness. I would then have subtraction of the CMB from the whole image to be under the control of a parameter.

I think the replacement of zero brightness with some arbitrary small value is an extremely ugly expedient. To the argument that division of images will fail if some pixels are zero I would answer that any proper division software (such as the examples of such I have written myself) would take account of the possibility of zero divisors and would replace such by NaNs in the output. The argument that image viewers cannot cope with NaNs I dismiss with similar contempt. However we live in a real world in which many sub-ideal things happen, including crap software being passed off as usable, and even becoming the standard. So excuse my cynicism but I think 1e-30 is probably here to stay.

tlunttil:
> No, because I haven't seen such things myself. What model did you use to see these effects?

This is basically from the example model from #206 with no CMB subtraction and 'outside the model' pixels set to CMB value (I'm using a slightly modified 1.7). Something seems to be going on near the edges of the model. It may not be important, but I'll have look.

ds9

allegroLeiden:
That's interesting because one of the changes I made between 1.6 and 1.7 was in the calculation of the number of extra rays to place on the circle. In the first pass of this algorithm I calculated the mean NN distance between **all** the projected <R rays and then generated circle rays with the same mean NN distance. However this averages all the projected grid points, and as there are a large number of these near the centre, the mean NN comes out much smaller than the NN value nearer the model edge, which resulted in a large number of circle points, and many thin, radially-oriented Delaunay triangles ending in circle points. The second version of the algorithm only included grid points in the outer third of the projected model circle. For 'normal' models this works out about right, but for Christian's #206 model the density is 0 outside the small cube, and I found that the number of points with r>2R/3 was also 0. This was a divisor in an expression leading eventually to the number of circle points. I made a small change in the present PR to catch such cases of zero divisor, but it should still result in 0 additional circle points, not of order 100 as you show above. So I don't really know what is going on there.

tlunttil:
It's possible that I made some special modifications that I forgot during the holidays and this problem may not even be present in official versions. Anyway, I'll see if I can find what's going on here.

allegroLeiden:
I thought I added the circle rays randomly but I just looked at the code and they are in fact evenly-spaced. Not sure why I chose this - simplicity? Quasi-random might be best.

I also found another instance of possible divide by 0, damn.

tlunttil:
It seems that some of the extra rays on the circle will because of numerical issues end up outside the model. Then in `traceray()` they bypass the normal CMB treatment and their intensity is always 0 (lines 122 and 123 in raytrace.c). This is probably not an issue in normal use, but without CMB subtraction it becomes visible.

allegroLeiden:
Ok... I did wonder about this issue at some stage. Probably the best belt-and-braces approach would be to (i) set the default for all rays/pixels to local_cmb; (ii) subtract local_cmb (if desired) from the whole image afterwards. I don't know if it is worth doing this mod.

tlunttil:
I agree. I would remove the subtraction from traceray/traceray_smooth and do it later with the interpolated image.

tlunttil:
Actually, I would move the whole CMB addition/subtraction thing to be done after computing the interpolated image and tau map of the model. It would be more correct to calculate CMB absorption using exp(-tau) with interpolated tau rather than interpolating exp(-tau) as is currently done.

allegroLeiden:
Interesting point. I'd hoped to merge this and move on to the next by now but you make a good case for some more mods.

allegroLeiden:
I'm going to merge it tomorrow if nobody objects. There's another bug fix waiting, then I want to do a patch release.

tlunttil:
When you add the 'circle rays' on the outer border of the projected model, due to numerical issues some of the rays will end up intersecting the model while others miss it. This means that not all rays are treated in the same way. It probably isn't a problem in practice, but it's IMO kinda unaesthetic ;)

I would consider adding on lines 876 and 877 of raytrace.c a safety factor `1.+epsilon` so the all added rays would be slightly outside the model (or maybe `1.-epsilon` so that they would be inside?).

allegroLeiden:
You're right, it is unaesthetic... what about simply declaring circle rays outside by fiat? Then there is no need to raytrace them at all, just set their intensity and tau to zero.

tlunttil:
Yes, that would be even cleaner and a bit faster, too.

tlunttil:
This looks good to me.

Issue #203, opened 2016-12-09
Cell-following raytrace algorithm is not thread-safe


imckstewart:
When par->traceRayAlgorithm==1 the software makes use of an algorithm which finds the chain of Delaunay cells traversed by each ray then interpolates quantities of interest inside the cells to solve the RTE. I note with some chagrin however that I neglected to make this procedure thread-safe, and at least one user has had a crash during raytracing in parallel when this option was chosen.

I'll put a note in the documentation for now.

tlunttil:
What breaks the thread-safety here?

allegroLeiden:
Errrr... maybe it is after all? Working far too hastily at the moment. F**k deadlines. F**k unrealistic expectations. Oh yeah, merry xmas everybody. Woohoo, kind of.

tlunttil:
I didn't immediately see anything, though I didn't look that carefully. I might take a closer look later today, but at least for the time being, the note in documentation should remain.

allegroLeiden:
Found this problem - it was a variable I declared static in extractFace(). I'll push a fix in a day or two.

allegroLeiden:
Closed by #208.

PR #208, opened 2017-01-18
Fixes the data race in raytracing (#203)

allegroLeiden:
I should make the caveat that the valgrind tool DRD still reports problems; however since these all occur outside the parallel section of raytrace.c, I don't understand what is going on. I attach the log file. Any ideas @tlunttil, or anybody else?

[dlog.txt](https://github.com/lime-rt/lime/files/713608/dlog.txt)

To generate this I did
```
  cd example
  ../lime -n -t -p 4 model_lte.c
  ^c
  valgrind --tool=drd --log-file=dlog.txt lime_<whatever>.x
```

BTW I was not able to squash the 'Minor' commit message via rebasing. Git seems to get confused after a merge.

### DIFF
--- a/src/aux.c
+++ b/src/aux.c
@@ -23,21 +23,6 @@ dotProduct3D(const double *vA, const double *vB){
 
 /*....................................................................*/
 void
-mallocInputPars(inputPars *par){
-  int i;
-
-  par->moldatfile = malloc(sizeof(char *)*MAX_NSPECIES);
-
-  par->collPartIds  = malloc(sizeof(int)*MAX_N_COLL_PART);
-  for(i=0;i<MAX_N_COLL_PART;i++) par->collPartIds[i] = 0;
-  par->nMolWeights  = malloc(sizeof(double)*MAX_N_COLL_PART);
-  for(i=0;i<MAX_N_COLL_PART;i++) par->nMolWeights[i] = -1.0;
-  par->dustWeights  = malloc(sizeof(double)*MAX_N_COLL_PART);
-  for(i=0;i<MAX_N_COLL_PART;i++) par->dustWeights[i] = -1.0;
-}
-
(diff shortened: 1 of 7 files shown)


tlunttil:
Maybe there are some dangling pointers from earlier parts which then confuse valgrind? Does memcheck complain about anything? On the other hand spurious error messages are not unheard of.

If you run with -DTEST a few times, are the results identical on bit-level? If so, there's probably no error.

BTW, it could be worth looking at parallelising the (unparallelised) regions that drd complained about. At least that interpolation part sometimes takes a non-negligible amount of time. (There's also a cosmetic issue that the progress bar/percentage doesn't take into account time spent there, so it may get stuck at near 1005 for a while.)

allegroLeiden:
> Does memcheck complain about anything?

No.

> If you run with -DTEST a few times, are the results identical on bit-level?

Yes.

> BTW, it could be worth looking at parallelising the (unparallelised) regions that drd complained about.

Maybe later - for now I just want to fix #203 and issue a patch.

allegroLeiden:
> Does memcheck complain about anything?

Oops - I looked too quick. There are zero errors, but there are some leaks. Probably unrelated to the drd problem, but I'll fix them anyway.

tlunttil:
Not related to drd warnings, but something I noticed when I was looking at the code: what is xySquared (starting from line 999 of raytrace.c) used for? Nothing, it seems...

allegroLeiden:
Hmm yeah, seems like fossilized code. I'll nix it.

tlunttil:
If the drd problems still remain, they could maybe be related to some dangling pointers confusing valgrind (it seems generally pointers are not NULLed in LIME after freeing). I might have a look at this later, but in my opinion you could merge this PR.

PR #209, opened 2017-01-23
Implemented 2nd-order interpolation of velocity during raytracing.

allegroLeiden:
For a while now I have wanted to increase the separation between the number-crunching, solving part of LIME and the raytracing part. Ideally I would like the user to be able to solve for a particular model, save all the solution data to file (FITS at present but I am working on an HDF5 interface), then at some later point to perform raytracing, just with the file as input, plus whatever image parameters are necessary. This is the most practical way to set up the CASA interface to LIME for example. In this outlook it is very helpful if the user does not have to supply model functions also for the raytracing bit. This has been generally ok with the exception of velocity. The problem with velocity has been that the line amplitude is sensitive to changes in velocity, and velocity can in fact change significantly across a Delaunay cell (at least in the example model that comes with the task). A linear interpolation of velocity within the cells was found not to provide adequate accuracy. The only other recourse has been to require the user to provide the velocity function both for the solution part of LIME and the raytracing part.

In the present PR I include code to perform a 2nd-order in-cell interpolation, which avoids the calls to velocity() in raytracing. The example image this produces is, it is true, noticeably different from that with no interpolation + calls to velocity(), but only at the 10% level. The spectra look ok. Thus I submit this PR for people to try out and comment on.

It would not be much extra work to increase the order to 3: we would need to keep 1 additional velocity sample at the barycentric centre of each cell, plus requiring the number of samples per edge to be an even number. Beyond that I would think it is probably pointless to go.


tlunttil:
I'll have a look at this.

tlunttil:
The last two parameters to `traceray_smooth` do not appear to be used anywhere.

Since the restriction of velocity along the ray inside a single Delaunay cell is a quadratic function, it would in principle be possible to do the integral (mostly) analytically and use a lookup-table in the same way as `geterf` does for linear velocity change and do away with substepping. It may be that the needed LUT would be impractically large, though.

allegroLeiden:
I had a quick look in Gradshteyn and Ryzhik but couldn't find any integral for exp(-(x^2+k)^2).

tlunttil:
Hmm, I don't think that's the integral that one would need... Anyway, after some further thinking I believe that the idea would be too messy to bother with at least at this time.

allegroLeiden:
> Hmm, I don't think that's the integral that one would need...

Which then? Just curious.

tlunttil:
It should be the convolution of a Gaussian with the distribution of projected velocity along the ray. Which works out to something that I wrote on a piece of paper which has been lost...

allegroLeiden:
I went into all the integral stuff in #153, won't repeat it, but I don't see how you can express that as a convolution. But anyway all this is fairly academic. The main question is whether you think the accuracy of the present interpolation is good enough to justify neglect of the velocity() function after the grid construction, which is the target I'm aiming for. I'd like to hear other opinions on it but everybody else seems to have gone quiet.

tlunttil:
Whether the method is accurate enough depends of course on the model. I'm sure that it is possible to construct (a very artificial) model where the quadratic interpolation is badly off. I doubt that would be the case with realistic models and even if that happens, it's also an indication that the grid wasn't fine enough to begin with.

So I'm fine with the quadratic interpolation.

allegroLeiden:
Do you think I ought to introduce, hmm well if not a parameter, then a macro in lime.h to control whether this is done or not?

tlunttil:
The non-smooth raytracing still calls velocity, right? I don't think there is any need to have a traceray_smooth version that doesn't use the quadratic interpolation.

allegroLeiden:
Duh I forgot I had already introduced a new img parameter to control it. I'll wait a couple of days before merging to see if 211 gets over the bar.

PR #211, opened 2017-01-24
Units-string-input

lolMEvans:
I have removed the need for multiple raytracings when only the image unit is changed between multiple image outputs. Now the user can set a delimiter (space, comma, colon or underscore all accepted) separated string of the units required, where the unit numbers-to-type are the same as before. The unit name and file extension are then automatically appended to img[].filename.

### DIFF
--- a/example/model.c
+++ b/example/model.c
@@ -116,14 +116,15 @@ input(inputPars *par, image *img){
    *        2:SI
    *        3:Lsun/pixel
    *        4:tau
-   * The unit name is then added automatically. If img[].filename does not have an extension then '.fits' will be
-   * appended automatically. Otherwise the existing extension is maintained.
+   * If multiple units are specified for a single set of image parameters (e.g. "0 1 2") then the unit name is added
+   * automatically at the end of the given filename, but before the filename extension if it exists. Otherwise if a
+   * single unit is specified then the filename is unchanged.
 
    * A single image unit can also be specified for each image using img[].unit as in previous LIME versions. Note that
    * only img[].units or img[].unit should be set for each image.
   */
 //  img[i].unit                   = 0;
-  img[i].units                  = "0 1";
+  img[i].units                  = "0 1 2";
   img[i].filename               = "image0";        // Output filename
 }
 
--- a/example/model.c
+++ b/example/model.c
@@ -108,6 +108,7 @@ input(inputPars *par, image *img){
   img[i].azimuth                = 0.0;
   img[i].incl                   = 0.0;
   img[i].posang                 = 0.0;
+
   /* For each set of image parameters above, numerous images with different units can be outputted. This is done by
    * setting img[].units to a delimited (space, comma, colon, underscore) string of the required outputs, where:
    *        0:Kelvin
@@ -115,10 +116,15 @@ input(inputPars *par, image *img){
    *        2:SI
    *        3:Lsun/pixel
    *        4:tau
-   * The unit name and file extension is then added automatically.
+   * The unit name is then added automatically. If img[].filename does not have an extension then '.fits' will be
+   * appended automatically. Otherwise the existing extension is maintained.
+
+   * A single image unit can also be specified for each image using img[].unit as in previous LIME versions. Note that
+   * only img[].units or img[].unit should be set for each image.
   */
+//  img[i].unit                   = 0;
   img[i].units                  = "0 1";
-  img[i].filename               = "image0";  // Output filename
+  img[i].filename               = "image0";        // Output filename
 }
 
 /******************************************************************************/
--- a/example/model.c
+++ b/example/model.c
@@ -105,11 +105,20 @@ input(inputPars *par, image *img){
   img[i].imgres                 = 0.1;            // Resolution in arc seconds
   img[i].distance               = 140*PC;         // source distance in m
   img[i].source_vel             = 0;              // source velocity in m/s
-  img[i].unit                   = 0;              // 0:Kelvin 1:Jansky/pixel 2:SI 3:Lsun/pixel 4:tau
-  img[i].filename               = "image0.fits";  // Output filename
   img[i].azimuth                = 0.0;
   img[i].incl                   = 0.0;
   img[i].posang                 = 0.0;
+  /* For each set of image parameters above, numerous images with different units can be outputted. This is done by
+   * setting img[].units to a delimited (space, comma, colon, underscore) string of the required outputs, where:
+   *        0:Kelvin
+   *        1:Jansky/pixel
+   *        2:SI
+   *        3:Lsun/pixel
+   *        4:tau
+   * The unit name and file extension is then added automatically.
+  */
+  img[i].units                  = "0 1";
+  img[i].filename               = "image0";  // Output filename
 }
 
 /******************************************************************************/
--- a/src/aux.c
+++ b/src/aux.c
@@ -48,7 +48,7 @@ The parameters visible to the user have now been strictly confined to members of
   double cos_pa,sin_pa,cosPhi,sinPhi,cos_incl,sin_incl,cosTheta,sinTheta,cos_az,sin_az;
   double tempRotMat[3][3],auxRotMat[3][3];
   int row,col;
-  char *pch_sep = " ,:_", *pch, *units_str;
+  char *pch_sep = " ,:_", *pch, *pch_end, *units_str;
 
   /* Check that the mandatory parameters now have 'sensible' settings (i.e., that they have been set at all). Raise an exception if not. */
   if (inpar.radius<=0){
@@ -165,8 +165,7 @@ The parameters visible to the user have now been strictly confined to members of
     /* Find out how many density functions we have (which sets par->numDensities).
     */
     for(i=0;i<MAX_N_COLL_PART;i++) dens[i] = -1.0;
-    density(0.0,0.0,0.0,dens); /* Note that the example density function in LIME-1.5 generated a singularity at r==0!
- * Such uglinesses should not be encouraged. I've fixed it now, thus I can use 0s here in (relative) safety. */
+    density(0.0,0.0,0.0,dens); /* Note that the example density function in LIME-1.5 generated a singularity at r==0! Such uglinesses should not be encouraged. I've fixed it now, thus I can use 0s here in (relative) safety. */
     i = 0;
     while(i<MAX_N_COLL_PART && dens[i]>=0) i++;
     par->numDensities = i;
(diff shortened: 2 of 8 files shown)


allegroLeiden:
My main issue with the changes is that they are non backward compatible. At some stage we may take the leap, have a good cleanup of the hoary old parameter interface, at which point we will discard BWC. But for the moment I think we ought to retain it where possible - and so far it has been possible - even where this necessitates some extra coding work.

In the present case I think you ought to:
  - Retain `unit` as an attribute of struct `image`. 
  - Add your `char*` attribute `units` to this struct, as you have already.
  - For struct `imageInfo`:
    * Delete attribute `unit`.
    * Add your `int` attributes `imgunits` and `numunits`. (Don't include these in struct `image` because the user doesn't set them directly.)
  - Set the default for `(*img)[i].unit` to 0 in `main.c:initParImg()`.
  - Inside the per-image loop in `aux.c:parseInput()` where you parse the `units` str, if `inimg[i].imgunits==NULL` (indicating the user hasn't set it) then copy the `inimg[i].unit` value to the first entry of `(*img)[i].imgunits` and set `(*img)[i].numunits` to 1. This retains BWC.

`aux.c` lines 169, 180, also in `defaults.c`: generally I don't think it is a good idea to make non-functional i.e. stylistic alterations to just a few bits of code in a multi-developer project. We all have our stylistic preferences, so really the only workable idea is to allow contributors of blocks of code, new functions for example, to express their own preferences within that block, but to leave others' alone elsewhere.

`aux.c` line 249, `main.c` lines 204,205,251,252: mixing tab and space indents in code is a bad idea because different editors translate a tab to different numbers of spaces.

`aux.c` line 259: I guess you are using `strtol()` here in case the user includes whitespace in their unit string but it is better then to cast it specifically to the `int` you are storing the value into.

`main.c:run()` changes:
  - A more explanatory name for `fits_filename` would be something like `img_filename_root`.
  - A more explanatory name for `fitsFilename()` would be something like `insertUnitStrInFilename()`.

`writefits.c:fitsFilename()`:
  - The argument `unit` is not the unit integer code, it is an index to the unit integer code. There is no charge for characters in variable names, a few more can remove potential confusion.
  - I am a bit uneasy about hard-wiring in the filename unit strings in this way, also the suffix `.fits`. Some users might prefer `.fit` or `.ds` or anything, really. I admit it is difficult to provide complete freedom without adding numerous extra knobs to twiddle (and LIME has too many of these already IMO), but as a compromise I would suggest that you allow the user to provide the whole filename still in attribute `filename` of struct `image`; then for the case where there is more than 1 unit in the string, just insert `_<unit int>` before the period, or if there is no period, tack `_<unit int>.fits` to the end of the supplied filename. All we want is some minimal scheme which takes only that amount of the user's freedom away necessary to add in the unit integers when it is necessary to do so to remove ambiguity. File names can be changed subsequently at will to include any string the user likes, and all the unit info is present in the file in the form of a keyword.

lolMEvans:
Thanks for the detailed comments @allegroLeiden. I'm new to working on a group project so I overlooked some things when I submitted this pull request, such as the code restyling of other authors' work and the importance of backwards compatibility. I've incorporated most of your suggestions into the branch and hope now that the pull request is more aligned with the existing version of LIME. Note, I've kept strtol in the units parsing block because I believe this is the safest way to convert a string to an integer, and I've now also added in a check to ensure the conversion is successful.

tlunttil:
Apropos stylistics and tab/space indents: maybe it would be useful to run the code through a beautifier such as uncrustify. I find the varying indent styles annoying and confusing and once I'm (soon I hope) done with the linepol stuff that part of code will bring yet another completely different style into the mix.

allegroLeiden:
I'm ok with unifying the styles at some point. Shall we pencil that in for after your linepol additions?

tlunttil:
> I'm ok with unifying the styles at some point. Shall we pencil that in for after your linepol additions?

Sounds good to me.

allegroLeiden:
Hi again @lolMEvans, I took a look at the revised code. Mainly I am still not happy that a user who says 'I want my image file named fred' will not in fact get fred but something like fred_Kelvin.fits. This is not a software issue, it is a user issue. Astronomers are wary, conservative animals and I can hear the howl of protest even now before it happens. Granted, if there are several units, then from the fact that there is only 1 filename supplied, ipso facto the only sensible solution is to take the filename as a root or template and add unit-specific parts to it. If you prefer these longer strings to integers then ok, but I want to see protection for the user's exact preference when there is only 1 unit, as will probably be the most common situation.

Software issues:

- Set the `image` struct element `units` to NULL after line 104 in main.c, otherwise the test for NULL at line 246 of `aux.c` will not work as desired.

- Will the test for NULL at line 259 of `aux.c` now ever return true? It worked before because `imgunits` was also an element of the `image` struct and was initialized to NULL in `main.c:initParImg()`, but there is now no explicit setting of `(*img)[i].imgunits` to NULL anywhere.

- I had no objection to the use of `strtol` as such, I agree it is the best way to convert the numbers. I just think it is better style to make casts explicit. I.e. I would change it to
```
        (*img)[i].imgunits[j-1] = (int)strtol(pch, &pch_end, 0);
```
at `aux.c` line 263.

- Now that you are calling `aux.c:copyInparStr()` from outside the module, include a template in `lime.h` to avoid compiler warnings.

- Same for `insertUnitStrInFilename()`.

lolMEvans:
Thanks again @allegroLeiden for your comments- I appreciate them as C is not exactly my forte. I have restored the verbatim filename behaviour if only one image unit is requested, and removed the default .fits extension so that the user has full control over the image filenames. I've also rectified the software issues.

> Will the test for NULL at line 259 of aux.c now ever return true? It worked before because imgunits was also an element of the image struct and was initialized to NULL in main.c:initParImg(), but there is now no explicit setting of (*img)[i].imgunits to NULL anywhere.

It is my understanding that (*img)[i].imgunits will return NULL if the realloc fails indicating that there was a memory error, hence the bail_out() message.

tlunttil:
Checking the return values of memory allocations is of course the proper thing to do. Currently LIME does very little of that, so in most cases of failed allocations it gracefully exits with seg fault ;)

allegroLeiden:
C is not my first language either. I was unaware for example that malloc or realloc returns NULL if it fails.

Valgrind still does not like the realloc though, I guess because at the first call, `(*img)[i].imgunits` is still undefined. If you set it to NULL just after line 243 of aux.c this error goes away.

If you make that change, and merge in the current master track, I'll merge your PR.

lolMEvans:
I suppose it's better to be explicit so that makes sense- I have made this change and merged the master branch.

allegroLeiden:
I didn't use any flags for valgrind, just the default:
```
ims@quitor:~/c_code/work/allegro/projects/lime/code_of_others/PR_211_c/example$ valgrind lime_31836.x
==32029== Memcheck, a memory error detector
==32029== Copyright (C) 2002-2015, and GNU GPL'd, by Julian Seward et al.
==32029== Using Valgrind-3.12.0 and LibVEX; rerun with -h for copyright info
==32029== Command: lime_31836.x
==32029== 
*** LIME, The versatile line modeling engine, version 1.7.1
>>> NOTE! Test flag is set in the Makefile. <<<
==32029== Conditional jump or move depends on uninitialised value(s)
==32029==    at 0x4C2FBF1: realloc (vg_replace_malloc.c:785)
==32029==    by 0x404D18: parseInput (aux.c:247)
==32029==    by 0x428795: run (main.c:176)
==32029==    by 0x402D96: main (main.c:291)
==32029== 
Warning : User didn't set par.dustWeights. Using the first density to calculate k_nu.
   Building grid: DONE                               
```
etc. I only set `unit` in the model file I used.

tlunttil:
Actually, why is line 248 using realloc instead of malloc?

allegroLeiden:
Good point. :)

tlunttil:
If null pointer is passed to realloc, it just mallocs. So it's not necessary to change it. (*img)[i].imgunits must be set to NULL anyway in the beginning, because otherwise realloc at line 261 doesn't work. If I understand the intention of this part of the code correctly, that realloc is needed. And while we're at it could the post-increment of at line 261 be taken out to a separate line?

allegroLeiden:
Curious that you make that last request. I would have said the j++ was quite a Lunttilism. ;)

@lolMEvans I'll just merge it as it is and make @tlunttil happy with a direct upload.

tlunttil:
Using pre/post-increments in anything but the simplest expressions feels like a great way to confuse everyone. In this case in the first iteration of while loop line 261 realloc is to size zero (which works out to be free() ). Is that the intended behaviour? Maybe it is, I haven't read the code further...

allegroLeiden:
@lolMEvans I just noticed a few things in writefits.c which I am puzzling about. A while ago I set up unit 5 as a sort of diagnostic: this wrote a 2D FITS image which just showed the number of rays per pixel. In `writeFits()` images of this unit are sent to `write2Dfits()`. There is no `scale` calculated or applied for this unit in `write2Dfits()`, nor is there any provision at all for images of unit==5 in `write3Dfits()`. However I see that you are calculating a `scale` value for unit==5 in `write2Dfits()` (but not applying it), and you have duplicated the 2D unit==5 stuff to 3D. Was this a mistake or did you have some other purpose in mind?

lolMEvans:
@allegroLeiden That is indeed a mistake that must have come about from some cross-contamination from a different branch I was working on when I was first using Git. Should I submit a pull request to restore write2DFits() and write3DFits() to their previous versions? Also, along these lines, why does writeFits2D() consider unit!=5 when it is only invoked when unit==5? Is it in case writeFits2D() will be used for other image units in the future?

> In this case in the first iteration of while loop line 261 realloc is to size zero (which works out to be free() ). Is that the intended behaviour?

@tlunttil This was not the intended behaviour and I see @imckstewart has thankfully already fixed this issue.

allegroLeiden:
Don't worry, I'll just fix and upload. That'll be quicker.

allegroLeiden:
> why does writeFits2D() consider unit!=5 when it is only invoked when unit==5?

`writeFits()` is in error here: I should amend it to invoke the 2D for continuum images too.

lolMEvans:
Ah ok I was assuming write3DFits() was being used for continuum as well, because nchan=1 in that case, and writeFits2D() was now only for number of rays per pixel. That change will certainly make the function names make sense but is there any need for two separate, almost identical, fits writing functions?

allegroLeiden:
Probably not (shrug). It seemed like a good idea at the time.

lolMEvans:
Fair enough- I'm not being critical I was just wondering if I'd missed something obvious because it wouldn't be the first time!

tlunttil:
So writeFits2D() is only invoked to output the number of rays per pixel, which is really a debug feature that isn't used by normal users, or continuum non-polarised images, for which LIME shouldn't be used anyway? Yeah, I agree that writeFits2D() could be removed. It's not that urgent, though.

lolMEvans:
> or continuum non-polarised images, for which LIME shouldn't be used anyway

@tlunttil I'm currently producing non-polarised continuum images for my model, with the intention to move onto line images shortly. Why do you say LIME shouldn't be used for non-polarised continuum images?

tlunttil:
I meant that if someone is only interested in dust emission images, there are much better tools available that can do the whole dust RT including the dust temperature calculation, not just the raytracing part.

lolMEvans:
Ah yea that's true- I have dust temperatures already so LIME fulfil my needs. Also, for that reason, I'd advocate keeping continuum output in some form, perhaps as it is currently in writeFits3D (renamed) whilst removing writeFits2D. I agree, though, that this is low priority.

Issue #201, opened 2016-12-08
Check for overflows in initial calls to model functions.


allegroLeiden:
See discussion in #198.

allegroLeiden:
Addressed by #214.

PR #214, opened 2017-02-09
Several changes to checks of mandatory functions.

allegroLeiden:
Checks made on the presence of the five 'mandatory' functions density(), temperature(), abundance(), doppler() and velocity(), are now performed more intelligently. A test for divide-by-zero is also done on the returns of these functions and a warning issued.

This addresses #201 and replaces (or invalidates) #198. Detailed changes are as follows:

  - In aux.c:parseInput() the presence of a par->gridInFile is tested, and if one has been supplied, the number of density columns is read from this and par->numDensities is set to this value.

  - In grid.c:readOrBuildGrid(), all tests for the presence of mandatory functions are now (also) done before smoothing. This is so, if one is missing, the user won't have to wait until the smoothing is done before finding out.

  - All tests of mandatory functions do so at the origin of coordinates. If the user has allowed a singularity at this location (for example if the function divides by some power of radius), a warning is issued. LIME should not be trying to second-guess where the user might have left a singularity: it should be the user's responsibility to avoid building such into their functions.

  - In grid.c:readOrBuildGrid(), if par->nLineImages==0, only the density() and temperature() functions are required when there are no file values.

  - New element doInterpolateVels of struct configInfo is set if any of the images has this flag set.

  - Restored free of gp[].w in freeSomeGridFields() which had been mistakenly deleted.

### DIFF
--- a/src/aux.c
+++ b/src/aux.c
@@ -14,6 +14,8 @@ TODO:
 #include <gsl/gsl_statistics.h>
 #include <float.h>
 
+#include "gridio.h" //**** should not need this - fix up these gridio macros!
+
 
 /*....................................................................*/
 double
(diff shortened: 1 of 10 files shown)


tlunttil:
I think that somewhere along the line predefgrid feature has been broken... It's probably enough to add if !predefgrid checks in a few places. Making traceray_smooth work with predefgrids would need a larger changes, but it's not that urgently needed anyway.

PR #218, opened 2017-02-13
Update copyright year

migueldvb:
Update of the copyright year in the source files.


allegroLeiden:
I've been trying to think of some way to automate this. Maybe a sed command in a configure script?

PR #224, opened 2017-04-12
Add fits extension to image file in example

migueldvb:


allegroLeiden:
Good idea, but shall we also go back to
```
  img[i].unit                   = 0;
//  img[i].units                  = "0 1 2";
```

PR #225, opened 2017-04-20
Quicker stateq

allegroLeiden:
In stateq.c the part of the matrix involving collision coefficients does not depend on the results of getjbar() and thus has been moved outside the loop now starting at line 141 (old line 45) in stateq.c. This results in a slight speedup.


tlunttil:
Late to this party, but couldn't you have done also the collision partner looping in getFixedMatrix? Why do you need to pass on (possibly) several collPartMat which are then summed together when you could've already done that in getFixedMatrix? Am I not seeing something?

allegroLeiden:
Good point, never thought of it. Damn you Tuomas, why do you always make me feel so dumb? :) We could also maybe replace the 'Initialize matrix with zeros' with initializing them with the collPart values.

tlunttil:
Maybe also take lines 90-96 out of getMatrix, as you commented in #177.

tlunttil:
Anyway, I'll try to cobble up something.

Issue #222, opened 2017-03-05
Problem finding qhull


kaifuski:
I have a problem when I tried to test the file "model.c" using LIME. It looked like how it would have a problem with qhull. Before I had installed libqhull-dev, version 2012.1-5, from debian repositories.

The first time I attempted run the program it showed an error like this.
```
/usr/bin/ld: no se puede encontrar -lqhullstatic
collect2: error: ld returned 1 exit status
Makefile:86: recipe for target '/home/jesus/Descargas/lime-master/lime_15305.x' failed
make: *** [/home/jesus/Descargas/lime-master/lime_15305.x] Error 1
```
The traduction of "no se puede encontrar" is "cannot find".
I tried to resolve the problem commenting the makefile in this way (lines 42,45,46,47):
```
#ifdef OLD_QHULL
     QHULL=qhull
     CPPFLAGS += -DOLD_QHULL
#else
#    QHULL=qhullstatic
#endif
```
With this, the program could run the first two steps (Building grid and  Smoothing grid) but then it showed this error.
`*** Error opening dust opacity data file!`
And also this.
```
jesus@asp14:~/Descargas/lime-master$ ./lime example/model.c
src/raytrace.c: In function ‘traceray_smooth’:
src/raytrace.c:733:29: warning: passing argument 2 of ‘doBaryInterpsVel’ from incompatible pointer type
       doBaryInterpsVel(DIM, vertexVels, edgeVels, edgeVertexIndices, entryCellBary, exitCellBary, doRay, rayVels);
                             ^
src/raytrace.c:382:6: note: expected ‘const double (*)[(sizetype)numDims]’ but argument is of type ‘double (*)[3]’
 void doBaryInterpsVel(const int numDims, const double vertexVels[][numDims], const double edgeVels[][numDims]\
      ^
src/raytrace.c:733:41: warning: passing argument 3 of ‘doBaryInterpsVel’ from incompatible pointer type
       doBaryInterpsVel(DIM, vertexVels, edgeVels, edgeVertexIndices, entryCellBary, exitCellBary, doRay, rayVels);
                                         ^
src/raytrace.c:382:6: note: expected ‘const double (*)[(sizetype)numDims]’ but argument is of type ‘double (*)[3]’
 void doBaryInterpsVel(const int numDims, const double vertexVels[][numDims], const double edgeVels[][numDims]\
      ^
[...]

PR #226, opened 2017-04-21
Slight speedup of raytracing.

allegroLeiden:
In traceray_smooth() the grid velocity is interpolated within the cell. This makes use of some buffer space. Previously this was being allocated at each call of the function and some unchanging indices were also calculated with each call. The buffer allocation and the unchanging calculation have now been moved outside the loop over rays.

Note that this also fixes the compiler warnings mentioned in #222.


tlunttil:
I don't know if there's any speed-up; the buffers in traceray_smooth() were variable-length arrays which are on the stack, so there was no mallocing. The new version is cleaner, though.

Issue #227, opened 2017-04-28
segfault when running Lime with electrons as only collision partner


gica3618:
I am trying to simulate (with version 1.7.2) carbon emission from a debris disk environment where electrons are the most important collision partners. I couldn't get it running, so I determined the minimum changes necessary to the model.c in the example folder that reproduce the error. The following two changes result in a segmentation fault:

- change par->moldatfile[0] to "c+.dat" or "c.dat"
- change par->collPartIds[0] to CP_e

I also changed img[i].trans to 0 (c+.dat contains only one transition), but anyway I think the error is produced before the image parameters become important.
When using CP_p_H2 instead of CP_e, no error occurs.
Is it possible to say why this error is produced?

tlunttil:
I get "Error: No recognized collision partners read from file.", not segfault. Anyway, it seems that reading the moldatafile goes off the rails if a coll.partner is found in the moldatafile, but not used by the model.

@allegroLeiden : if block starting line 150 of molinit.c is only run if the current coll.partner is requested by the model. But if it's not run, the file reading goes completely out of sync. I think fscanf etc. around lines 188-217 should always be done (and results discarded if they're not needed) to ensure proper reading of LAMDA molecule data files.

tlunttil:
@gica3618 : As a quick fix, you can edit the moldata file so that the collisional transition data with electrons comes as the first coll. data in the file; currently the first partner is read correctly.
[c+.dat.txt](https://github.com/lime-rt/lime/files/965179/c.dat.txt)

allegroLeiden:
I'll have a look.

tlunttil:
I could take a look at fixing this myself, tomorrow I'll be spending a couple of hours on a train with nothing much to do. Or do you have some PRs that will have an effect on this?

allegroLeiden:
There are a few PRs and proto-PRs pending but nothing affecting molinit.c. Knock yourself out. :)

allegroLeiden:
[...]

PR #228, opened 2017-05-03
Add a call to readDummyCollPart to keep file reading in sync.

tlunttil:
Simple fix to make sure that file reading doesn't go off the rails when there are unused collision partners defined in the molecule data file.

### DIFF
--- a/src/molinit.c
+++ b/src/molinit.c
@@ -217,9 +217,10 @@ void readMolData(configInfo *par, molData *md, int **allUniqueCollPartIds\
             fscanf(fp,"\n");
           }
         } /* End if(par->lte_only) */
-
         k++;
-      } /* End if CP found in par->collPartIds. */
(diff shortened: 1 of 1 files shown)


tlunttil:
Simple fix to #227, that is.

allegroLeiden:
Dumb error, simple fix. I think what I'll do with this is upload the changed lines directly to the lime-rt 'release' branch, release the patch as LIME-1.7.3, then finally merge this to 'master'. I think that would be the git-approved way. It would mean rejecting this PR though. :)

tlunttil:
Yep, I guess it's better that way.

I honestly thought there'd be a bit more work needed when I volunteered to to fix #227...

allegroLeiden:
Sure, sure. :)

I added a couple more elementary checks.

allegroLeiden:
Uploaded the fix direct to release_track branch and merged to master.

PR #230, opened 2017-05-11
Faster stateq

tlunttil:
Some rearrangement in stateq.c to make things faster and cleaner. Also removed some old unused code.

### DIFF
--- a/src/aux.c
+++ b/src/aux.c
@@ -878,7 +878,7 @@ While this is off however, other gsl_* etc calls will not exit if they encounter
         double *halfFirstDs;	// and included them in private() I guess.
         mp=malloc(sizeof(gridPointData)*par->nSpecies);
 
-#pragma omp for schedule(dynamic)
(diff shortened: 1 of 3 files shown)


tlunttil:
While I was testing this PR I noticed that with >1 threads results from different runs are not bit-level identical even if -t switch is used. This was not caused by this PR, also master branch has the same issue. I'll have a look tomorrow.

tlunttil:
The problem was dynamic omp scheduling in levelPops(). I removed it and now the results from different runs are identical when -t switch is used. There may be a slight performance penalty, but it's probably not enough to make any difference.

allegroLeiden:
Thanks for this, I'll have a look at it next week.

allegroLeiden:
Sorry I was so long checking this. Our bloody stupid story of disk failures which started last Nov continues unabated, it's been keeping me busy. :(

Issue #231, opened 2017-05-12
Can't write with par->gridOutFiles[i]


richteague:
When I try and export the grids with `par->gridOutFiles[i]` I encounter the error `Keyword 0 dataype 82 is not currently accepted.`. The models run through properly and the output is as expected so I think this is just an issue with the writing of the grids. It happens for gridding at all stages.

tlunttil:
@allegroLeiden : There seems to be a conflict between datatypes defined in gridio.h and fits datatypes (from fitsio.h); datatype 82 is fits TDOUBLE while gridio.h uses datatype 3 for lime_DOUBLE. I'm going to leave this one to you, since you know what the original intention was.

The typo in error message could also be fixed...

allegroLeiden:
Yeah I actually have this fixed in a branch which I will merge after your PR. I guess I ought to see if 1.7.3 has the same problem and patch that separately.

allegroLeiden:
Ok 1.7.3 does not have this problem so there is no need for a patch. As soon as I can merge the fixed code into the master branch it will go away from there as well. No action until then.

allegroLeiden:
PR #233 has closed this in master branch.

PR #233, opened 2017-05-18
Zero images

allegroLeiden:
This allows the user to run LIME with zero images specified, and thus introduces the facility to run LIME in two separate sessions (desirable for CASA integration): the first to solve the RTE and save the results to file, the second to read the file and create raytraced images from it. A new parameter `par.doSolveRTE` has been added to compensate for the loss of information when zero images are specified.


allegroLeiden:
This also fixes #231.

PR #235, opened 2017-05-29
Adds capability to read and write grid data to HDF5.

allegroLeiden:
Previously the only file format available has been FITS. Now HDF5 can be selected. Currently this must be done by hand via commenting/uncommenting appropriate lines of gridio.h. Better would be to do it via a Makefile switch.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,6 @@ before_install:
   - sudo add-apt-repository ppa:ubuntu-toolchain-r/test -y
   - sudo apt-get update
 install:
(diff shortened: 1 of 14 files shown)


tlunttil:
I'm afraid I won't have time to go through all 2000+ lines of new code... However, your truly Stakhanovite effort is highly appreciated.

Is there a specific reason why you're not using a Makefile switch for selecting the file format? It might also be better to set the default to using just fits. Not all systems have HDF5 libraries installed.

allegroLeiden:
Thanks for the kudos. :) No, I was going to add a Makefile switch before I merged. It was pure laziness not to do so before. Point about fits/hdf5 also well taken.

PR #239, opened 2017-07-03
Extensive code rearrangement.

allegroLeiden:
I've moved a lot of functions (and some struct definitions) around to try to simplify and slim down lime.h. I am also looking forward to introducing a python interface as well as a grid-file format translator, both of which will require some rearrangements.

There are new modules solver.c, grid_aux.c and run.c; I have deleted LTEsolution.c, photon.c, stateq.c and velospline.c. Unused modules have been moved to a subdirectory of that name.

This change will be difficult to check because diff will show huge differences, but the _only_ changes to code are the following:

  - Replaced 'typedef struct {blah} point;' in lime.h with 'struct point {blah};'

  - Replaced 'typedef struct {blah} spec;' in lime.h with 'struct spec {blah};'

  - In the 2 or 3 places where it occurred, replaced `sizeof(point)` and `sizeof(spec)` in mallocs with `sizeof(*<name of the pointer>)`

  - The function run() now returns an integer status. Currently this value is untested, and the only value returned is zero.

@rschaaf-aifa how are these changes in the ARTIST context? Good, bad or indifferent?


rschaaf-aifa:
The changes seem to be indifferent from ARTIST's perspective.

tlunttil:
Looks good, some serious rearrangement and clean-up was really needed.

A few comments:
- Is there a need to keep the 'unused' source files? Besides being unused, they're unlikely to be compatible with current version of LIME. Should their functionality be needed, the old code can be dug up from old versions on github.  
- There'd probably be a non-negligible performance gain from inlining some functions (dotProduct3D and gaussline at least). That's unfortunately a bit hard to do, because some older compiler versions (which probably still need to be supported) don't necessarily support it.
- A long-standing peeve of mine: 'photon' is an almost maximally uninformative function name. Maybe change it to 'calculateJBar' and change 'getjbar' to 'updateJBar'. 'stateq' and perhaps even 'levelpops' could also be renamed to be a bit more descriptive.
- Another thing, which isn't exactly related to this PR: should something be done to get rid of the compiler warnings for ignoring return value of fscanf in various places?

By the way, do the statistics min and median (SNR) calculated during the main levelpops loop tell anything relevant currently?

allegroLeiden:
I'll get rid of the 'unused_code' but the other points, while all good, I think should be left for other PRs.

PR #240, opened 2017-07-05
Cleanup

tlunttil:
General code cleanup. Added a lot of asserts to eliminate compiler warnings. Renamed some functions.

The default number of smoothing iterations when using  par->samplingAlgorithm=0 was changed from 20 to 5.


allegroLeiden:
Seems fine. Thanks for all the hard work adding 'asserts'! Just 1 or 2 things:

  - raytrace.c:295 - 'calculateJBar' omitted.

  - raytrace.c:1250 - is it worth a mention somewhere that this value actually changes now from 2.728 to 2.725? Possibly not?

tlunttil:
I'm amazed that you spotted the first one...

For CMB value, I'm not sure where that 2.728 came from; the correct value is 2.725 (2.72548±0.00057 K to be exact...), which was also used for excitation calculation. I sort of assumed that it was a typo in raytrace. A change of 3 mK isn't going to make any noticeable difference anyway.

tlunttil:
Using asserts like this in production code is of course not to everyone's liking... But I really don't feel like writing proper error handling code right now and I guess this is still better than just trying to keep going even if some file read fails.

PR #242, opened 2017-07-13
Python model files & configure script

### LINKED ISSUE #222: Problem finding qhull

allegroLeiden:
This should supersede #33 and fix #222.

LIME until now has required the user to write a model file in C, which is not among the most easy or popular languages. With this PR, additional make targets and code are provided which allow the user to write their model file in python. A template file `model.py` is provided in `examples/`.

The second advance in the present PR is to include a rudimentary configure script. This should locate the libraries and include files for qhull, cfitsio and python which are appropriate to the platform on which LIME is run.


tlunttil:
Unless I'm reading gcc docs somehow very wrong, the switch is -Wno-unused-result, not -Wno-unused-value...

tlunttil:
Minor comments, not sure if I'll have time to read the code well enough to say anything substantial:

- I'd put parentheses around the ARCSEC_TO_RN expression. And wouldn't RAD be clearer than RN?
- It seems to me that oneOver_i stuff could also be moved inside #ifdef FASTEXP, just as EXPTABLE.
- It looks like the new config would fail in what is the old default: when the qhull library is libqhullstatic.a. You seem to assume that it's libqhull. I don't know if such installations exist...

allegroLeiden:
- I started getting an warning with -Wno-unused-result, googled, and came up with comments like [https://stackoverflow.com/questions/19464892/cc1plus-wno-unused-result-error](https://stackoverflow.com/questions/19464892/cc1plus-wno-unused-result-error) or [https://stackoverflow.com/questions/19464892/cc1plus-wno-unused-result-error](http://lists.mindrot.org/pipermail/openssh-bugs/2011-May/010341.html). I can't find mention of either alternative in [https://gcc.gnu.org/onlinedocs/gcc-2.95.2/gcc_2.html#SEC8](https://gcc.gnu.org/onlinedocs/gcc-2.95.2/gcc_2.html#SEC8)

- libqhullstatic.a: This is not available on any of the platforms I can access, and I thought we agreed that it was never going to be appropriate? Your mail on the 4th.

FYI I am leaving to start 2 weeks holidays tomorrow so this PR will probably sit inactive until I get back. I had hoped to add command-line options but the multi-threaded stuff is proving tricky and I doubt I will have time to finish it today.

tlunttil:
- Docs for more recent gcc versions (at least 4.8 and more recent) do mention -Wno-unused-result. Which version of gcc you were using?
- I've never seen libqhullstatic anywhere either, so I'm fine with removing that option. If some platform does use a library called libqhullstatic, the user can always rise an issue...

allegroLeiden:
That may be the issue... our desktop machines running fedora-21 have gcc-6.3.1 but the RH machines on which we have our lustre system and run CASA have only gcc-4.7.7. Indeed the 6.3.1 version seems to have both flags, 4.7 only unused-value.

allegroLeiden:
I just saw that I looked up really old gcc pages. Sorry, it is Monday, haven't woken up yet.

tlunttil:
Maybe use -Wall, so we get all compiler warnings?

allegroLeiden:
I should add that this PR supersedes (at last) #33.

PR #249, opened 2017-09-11
Fixes #248

### LINKED ISSUE #248: Seg fault when par->nSolveIters left at 0.
The array `gp[].mol[].pops` is malloc'd under only 3 circumstances: if the user reads the pop info from a file (either via `par->restart` or `par->gridInFile`) or `par->lte_only` or `par->nSolveIters>par->nSolveItersDone`. If the user doesn't do any of these, but `par->doMolCalcs` still gets set, LIME seg faults due to trying to read `gp[].mol[].pops` without it being malloc'd.

allegroLeiden:
Added some code to exit with an error in the situation which previously generated a seg fault.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,6 @@ before_install:
   - sudo add-apt-repository ppa:ubuntu-toolchain-r/test -y
   - sudo apt-get update
 install:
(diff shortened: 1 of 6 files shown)


allegroLeiden:
Superseded by #250.

PR #254, opened 2017-12-28
Modify colli matrix population to avoid transposing

migueldvb:
This avoids the need to transpose the colli matrix as defined in `getFixedMatrix`.


allegroLeiden:
Hi Miguel, sorry not to reply before. I'll have a look and get back to you asap.

